PR #2, opened 2022-01-31
Solution

Abnormaltype:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe):
```
IvanRamyk, line 5: This function returns "Friends should buy 1 masks" istead of "All friends should be vaccinated"
for `[not_wearing_mask, not_vaccinated]` case

IvanRamyk:
```
(
    [
        {
            "name": "Alisa",
            "vaccine": {
                "name": "Pfizer",
            },
            "wearing_a_mask": False,
        },
        {
            "name": "Bob",
            "wearing_a_mask": True,
        },
    ],
    Cafe("KFC"),
),
```

app/main.py
```diff
@@ -1 +1,18 @@
...
+    masks_to_buy = sum([
```
IvanRamyk, line 6: Do not create a list to calculate a sum of generator elements
Use just `sum(1 for ....)`

IvanRamyk (changes requested):

app/main.py
Abnormaltype: okay, thx

app/main.py
Abnormaltype: I've fixed

app/main.py
```diff
@@ -1 +1,20 @@
...
+
+
+def go_to_cafe(friends: list, cafe: Cafe):
+    masks_to_buy = 0
+    is_vaccine = True
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            is_vaccine = False
```
IvanRamyk, line 12: You can return `'All friends should be vaccinated'` message here and not use `is_vaccine` variable
But, this solution is ok anyway)

IvanRamyk (approved):
Good job!

app/main.py
Abnormaltype: okay, thx!

PR #3, opened 2022-01-31
'Solution'

cloudprogramm:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe):
+    masks_to_buy = sum([
+        1
+        for friend in friends
+        if not friend["wearing_a_mask"] or "wearing_a_mask" not in friend
```
Krimat, line 9: There you check if visitor wear mask twice. First time in go_to_cafe function, second time inside visit cafe function

Krimat (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+    OutdatedVaccineError
+
+
+class Cafe:
+    def __init__(self, name: str):
+        self.name = name
+        self.today = datetime.date.today()
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" in visitor.keys():
```
IvanRamyk, line 13: You can simplify this function and write without nested ifs.
To do this, check first if the "vaccine" key is **not** present in the dict.
```
if "vaccine" not in visitor.keys():
     raise NotVaccinatedError("Some message")
```

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+            if visitor["vaccine"]["expiration_date"] < self.today:
+                raise OutdatedVaccineError("OutdatedVaccineError")
+            elif not visitor["wearing_a_mask"]:
+                raise NotWearingMaskError("NotWearingMaskError")
+            else:
```
IvanRamyk, line 19: `else` is unnecessary here

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, name: str):
+        self.name = name
+        self.today = datetime.date.today()
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" in visitor.keys():
+
+            if visitor["vaccine"]["expiration_date"] < self.today:
+                raise OutdatedVaccineError("OutdatedVaccineError")
+            elif not visitor["wearing_a_mask"]:
```
IvanRamyk, line 17: Use `if` instead of `elif` in such cases

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+class Cafe:
+    def __init__(self, name: str):
+        self.name = name
+        self.today = datetime.date.today()
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" in visitor.keys():
+
+            if visitor["vaccine"]["expiration_date"] < self.today:
+                raise OutdatedVaccineError("OutdatedVaccineError")
```
IvanRamyk, line 16: "OutdatedVaccineError" is not a descriptive message. You should add some information about an error in the corresponding message.

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #4, opened 2022-01-31
Solution

TPolina:


Oleksandr563487953 (approved):

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def visit_cafe(self, visitor: dict):
+        if 'vaccine' not in visitor:
+            raise NotVaccinatedError(f"{visitor['name']} is not vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError(
+                f"{visitor['name']}'s vaccine is expired")
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError(f"{visitor['name']} is not wearing mask")
+        else:
```
IvanRamyk, line 21: `else` is unnecessary here

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe):
```
IvanRamyk, line 5: This function returns "Friends should buy 1 masks" instead of "All friends should be vaccinated"
```
(
    [
        {
            "name": "Alisa",
            "vaccine": {
                "name": "Pfizer",
            },
            "wearing_a_mask": False,
        },
        {
            "name": "Bob",
            "wearing_a_mask": True,
        },
    ],
    Cafe("KFC"),
),
```

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #5, opened 2022-01-31
'Solution'

liuda-chuzdiuk:


app/errors.py
```diff
@@ -0,0 +1,17 @@
+class VaccineError(Exception):
+    pass
+
+
+class NotVaccinatedError(VaccineError):
+    def __init__(self, message="NotVaccinatedError"):
```
IvanRamyk, line 6: It's a bit strange message because you just copied the name of the error. 
It will be shown for user like this: 
```
NotVaccinatedError: NotVaccinatedError
```
It'd better to explain when and why this error is raised

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe):
```
IvanRamyk, line 4: This function returns "Friends should buy 1 masks" instead of "All friends should be vaccinated"

IvanRamyk:
```
(
    [
        {
            "name": "Alisa",
            "vaccine": {
                "name": "Pfizer",
            },
            "wearing_a_mask": False,
        },
        {
            "name": "Bob",
            "wearing_a_mask": True,
        },
    ],
    Cafe("KFC"),
),
```

IvanRamyk (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+    OutdatedVaccineError, NotWearingMaskError
+
+
+class Cafe:
+    def __init__(self, name):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
```
IvanRamyk, line 12: Could you add messages?)

IvanRamyk (changes requested):

app/cafe.py
liuda-chuzdiuk: i added messages in errors.py in __init__

IvanRamyk (approved):
Good job!

PR #6, opened 2022-01-31
'Solution'

alexandra-halytska:


Krimat (approved):

app/cafe.py
```diff
@@ -0,0 +1,17 @@
...
+    OutdatedVaccineError, NotWearingMaskError
+
+
+class Cafe:
+    def __init__(self, name):
+        self.name = name
+
+    def visit_cafe(self, visitor):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
```
IvanRamyk, line 12: Please, add descriptive messages for errors

IvanRamyk: This function returns "Friends should buy 1 masks" instead of "All friends should be vaccinated"

IvanRamyk:
```
(
    [
        {
            "name": "Alisa",
            "vaccine": {
                "name": "Pfizer",
            },
            "wearing_a_mask": False,
        },
        {
            "name": "Bob",
            "wearing_a_mask": True,
        },
    ],
    Cafe("KFC"),
),
```

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!!

PR #7, opened 2022-01-31
Solution

Krimat:


app/main.py
```diff
@@ -1 +1,20 @@
...
+
+
+def go_to_cafe(cafe: Cafe, friends: list):
+    mask_to_buy = 0
+    vacs_group = True
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except err.VaccineError:
+            vacs_group = False
```
denyslipin, line 12: In the task it is mentioned "If at least one of the friend has problems with vaccines the function should return a message `"All friends should be vaccinated"." So, you don't need to create vacs_group = True, then change if to False and then make verification the line 16-17, you just need to find out first such person and return the message

Krimat: I agree that in this task it`s possible to do such thing, but in general this will produce "spaghetti" code. Also, this code doesn`t meant to be as fast as possible, so it`s allowed to go through whole list every time.

denyslipin (approved):

app/cafe.py
```diff
@@ -0,0 +1,26 @@
+import datetime
+
+import app.errors as errors
+
+
+class Cafe:
+    def __init__(self, name):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
```
IvanRamyk, line 10: It's a complicated function to read
You use a lot of nested ifs
You can check requirements one-by-one

app/main.py
IvanRamyk: I think the suggestion above won't really lead to the complicated code. It can be even more readable because you will show explicitly that` VaccineError` lead to a specific returned value

IvanRamyk (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+class Cafe:
+    def __init__(self, name):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if 'vaccine' not in visitor:
+            raise errors.NotVaccinatedError(
+                f"Visitor {visitor['name']} has no vaccine"
+            )
+        if not visitor['vaccine']['expiration_date'] >= datetime.date.today():
```
IvanRamyk, line 15: You can write `val1 < val2` instead of `not val1 >= val2`
It'd be more readable

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #9, opened 2022-01-31
Solution

Nazar-kob:


app/main.py
```diff
@@ -1 +1,18 @@
...
+
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    return f"Friends should buy {masks_to_buy} masks" \
+        if masks_to_buy \
```
IvanRamyk, line 17: This condition works, but it'd be better to add explicit comparison with zero to increase code readability.
Also, using regular `if`s might be better than the ternary operator here.

IvanRamyk (changes requested):

app/main.py
Nazar-kob: Thanks, I'll fix it

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
+        return f"Friends should buy {masks_to_buy} masks"
+    else:
```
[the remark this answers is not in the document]
Nazar-kob: I got it, ok

IvanRamyk (approved):

PR #10, opened 2022-01-31
Solution

vldtsw:


Oleksandr563487953 (approved):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends, cafe):
```
IvanRamyk, line 4: This function returns "Friends should buy 1 masks" instead of "All friends should be vaccinated"
```
(
    [
        {
            "name": "Alisa",
            "vaccine": {
                "name": "Pfizer",
            },
            "wearing_a_mask": False,
        },
        {
            "name": "Bob",
            "wearing_a_mask": True,
        },
    ],
    Cafe("KFC"),
),
```

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
...
+    without_masks = sum(not friend["wearing_a_mask"] for friend in friends)
+    try:
```
IvanRamyk, line 6: Nested exceptions are difficult to read (
Please, try to simplify your code
If you need help to do this, feel free to ask me on QnA

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def visit_cafe(self, visitor):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(f"{visitor['name']} is not vaccinated")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError(f"{visitor['name']}'s "
+                                       f"vaccine is expired")
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError(f"{visitor['name']} is not wearing mask")
+        else:
```
IvanRamyk, line 20: This `else` is unnecessary here

IvanRamyk (changes requested):

app/cafe.py
vldtsw: Fixed

app/main.py
vldtsw: Thanks! I think fixed it

IvanRamyk (approved):
Good job!

PR #11, opened 2022-01-31
develop

Oleksandr563487953:


serhii56465 (approved):

app/cafe.py
```diff
@@ -0,0 +1,30 @@
...
+    OutdatedVaccineError
+
+
+class Cafe:
+    def __init__(self, name):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("NotVaccinatedError")
```
IvanRamyk, line 24: Please, add more descriptive messages for errors. Messages like this will not show any extra information about an error but just a copy of the error name
It'll look like this:
```
NotVaccinatedError:  NotVaccinatedError
```

app/cafe.py
```diff
@@ -0,0 +1,30 @@
+"""
```
IvanRamyk, line 1: Good job!

app/main.py
```diff
@@ -1 +1,27 @@
...
+def go_to_cafe(friends, cafe):
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy:
```
IvanRamyk, line 25: This condition works, but it'd be better to add an explicit comparison with zero to increase the code readability

IvanRamyk (changes requested):

app/cafe.py
Oleksandr563487953: Changes provided!

app/main.py
Oleksandr563487953: Changes provided!

IvanRamyk (approved):
Good job!

PR #14, opened 2022-01-31
Solution

Geksus:


Oleksandr563487953 (approved):

app/errors.py
```diff
@@ -0,0 +1,31 @@
+"""
+This module holds possible errors
+"""
+
+
+class NotWearingMaskError(Exception):
+    """
+    No mask -> virus gets in -> die
```
IvanRamyk, line 8: I got the joke :)
But code should be readable and clear, so it's not a good idea to write jokes in the docs. It's better to write descriptive and informative messages.
You can share your jokes in the chats, it would be better :)

Could you rewrite or remove such docs, please?)
Btw, it's nice that you decided to write docstrings, go on!

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends, cafe: Cafe):
```
IvanRamyk, line 5: Your function won't pass this test:
```
(
    [
        {
            "name": "Alisa",
            "vaccine": {
                "name": "Pfizer",
            },
            "wearing_a_mask": False,
        },
        {
            "name": "Bob",
            "wearing_a_mask": True,
        },
    ],
    Cafe("KFC"),
),
```
I described this case in our slack chat, you can check it there

>Often, if an error occurs that one of the visitors is not wearing a mask, you will immediately return a message that several friends should buy a mask for themselves.
But this is wrong because if one of the following friends is not vaccinated, you need to return an error about vaccination.

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #16, opened 2022-02-01
'Solution'

serhii56465:


app/main.py
```diff
@@ -1 +1,17 @@
...
+            try:
+                cafe.visit_cafe(visitor)
+            except NotWearingMaskError:
+                masks_to_buy += 1
+        if masks_to_buy == 0:
+            return f"Friends can go to {cafe.name}"
+    except VaccineError:
+        return "All friends should be vaccinated"
+    else:
+        return f"Friends should buy {masks_to_buy} masks"
```
IgorOlifirenko, line 17: You can use one try / except block

IgorOlifirenko (changes requested):

IgorOlifirenko (approved):

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Cafe:
+    def __init__(self, name):
+        self.name = name
+
+    def visit_cafe(self, visitor):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("You should be vaccinated")
+        today = datetime.date.today()
+        date1 = datetime.date(today.year, today.month, today.day)
```
IvanRamyk, line 16: Why do you copy `today`. Just use _it_

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+        if visitor["vaccine"]["expiration_date"] != datetime.date.today():
```
IvanRamyk, line 17: Why do you call `datetime.date.today()` one more time?

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+            date2 = visitor["vaccine"]["expiration_date"]
+            if date1 > date2:
```
IvanRamyk, line 19: You can use one `if` instead of `if`s in lines 17 and 19

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #20, opened 2022-02-01
Solution

anitaposliedova:


Abnormaltype (approved):
Good job!

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+                        NotWearingMaskError)
+
+
+class Cafe:
+
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if 'vaccine' not in visitor:
```
IvanRamyk, line 13: It's better to use the same quote style throughout the project
I suggest using double-quotes. Please, correct it :)

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #22, opened 2022-04-02
Solution . Katiukha

KatiukhaO:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from app.errors import NotVaccinatedError, \
+    NotWearingMaskError, \
+    OutdatedVaccineError
```
MasterpieceElbow, line 3: Bad code style, better to use
```
from module import (
    Class_1,
    Class_2,
    ...
)
```

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, \
+    NotWearingMaskError
```
MasterpieceElbow, line 3: Bad code style here too.

MasterpieceElbow (changes requested):

PR #26, opened 2022-05-26
Solution

YuriiKindrat:


rotsen18 (changes requested):

rotsen18 (approved):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe):
+    mask_to_buy = 0
+    for people in friends:
```
MasterpieceElbow, line 6: `people` means plural, while you get a single object by each iteration. Rename.

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe):
```
TPolina, line 4: if you add type annotation, add it to cafe as well

TPolina (changes requested):

TPolina (approved):
Nice!

PR #29, opened 2022-05-26
Solution

Serhii-Leonenko:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+                        OutdatedVaccineError)
+
+
+class Cafe:
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
```
rotsen18, line 13: you should pass descriptive messages when you raise the exception in the visit_cafe method

rotsen18 (changes requested):

app/cafe.py
Serhii-Leonenko: > you should pass descriptive messages when you raise the exception in the visit_cafe method

But in the task w don't have information about description of the error. Or I make a mistake?

rotsen18: we have)
https://prnt.sc/co7RFtR2fUeL

Serhii-Leonenko: aaa, ok.
Because, I looked on examples only, thanks

rotsen18 (approved):

TPolina (approved):
Good job!

PR #31, opened 2022-05-26
All done.

Flashmobber:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+import datetime
```
rotsen18, line 4: correct way to import modules:
1. Standard library imports
2. Third-party imports
3. Application-specific imports

rotsen18 (approved):

TPolina (approved):
Good job!

PR #32, opened 2022-05-27
'Solution'

DmytroSvirsa:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
```
rotsen18, line 16: you should pass descriptive messages when you raise the exception in the visit_cafe method

rotsen18 (changes requested):

app/cafe.py
DmytroSvirsa: where have you read that?
image

rotsen18 (approved):
Good job

TPolina (approved):

PR #33, opened 2022-05-27
Solution

RomaTsybulia:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+from app.errors import \
+    NotWearingMaskError, \
+    OutdatedVaccineError, \
+    NotVaccinatedError
```
rotsen18, line 5: better way is to use brackets

app/main.py
```diff
@@ -1 +1,16 @@
...
+def go_to_cafe(friends: list, cafe):
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy != 0:
```
rotsen18, line 13: just write 
```suggestion
    if masks_to_buy:
```

app/main.py
```diff
@@ -1 +1,16 @@
...
+        return f"Friends should buy {masks_to_buy} masks"
+    else:
```
rotsen18, line 15: you don't need else here

rotsen18 (changes requested):

rotsen18 (approved):

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+)
+
+
+class Cafe:
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
```
TPolina, line 15: an error message is strange. You perform checks for one visitor, so if he is not vaccinated why the message is "all friends". Who are these friends?

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError
```
TPolina, line 17: add descriptive error message

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("Friends should buy 2 masks")
```
TPolina, line 19: same here: your method receives one visitor but the error message is "Friends should buy 2 masks".
here you have to write your own error messages, not copy them from `go_to_cafe` function and example

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe):
```
TPolina, line 4: if you add type annotation, add it to cafe as well

TPolina (changes requested):

TPolina (approved):
Well done!

PR #36, opened 2022-05-28
'Solution'

SergeyDev1996:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        if visitor["vaccine"]["expiration_date"]\
+                < datetime.date.today():
+            raise OutdatedVaccineError
+        if "wearing_a_mask" not in visitor or\
+                visitor["wearing_a_mask"] is False:
```
Magus6784, line 20: if not visitor["wearing_a_mask"]:
        
       Work and cleaner

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+)
+
+
+class Cafe:
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
```
Magus6784, line 15: Add "error message" to your errors

app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+import datetime
+
+from app.errors import NotVaccinatedError, NotWearingMaskError
+
+
+def go_to_cafe(friends, cafe):
```
Magus6784, line 6: annotation is love

app/main.py
```diff
@@ -1 +1,25 @@
...
+            if "vaccine" not in friend or \
+                    friend['vaccine']['expiration_date']\
+                    < datetime.date.today():
+                raise NotVaccinatedError("All friends should be vaccinated")
+        except NotVaccinatedError:
+            return "All friends should be vaccinated"
+    for friend in friends:
+        try:
+            if friend['wearing_a_mask'] is False:
+                raise NotWearingMaskError
```
Magus6784, line 22: You already did this in class Cafe
You must just use method visit_cafe and catch errors

Magus6784 (changes requested):

app/main.py
SergeyDev1996: it won't me allow me to add the annotation for cafe, not sure why. https://prnt.sc/SVI0RiUTqRns

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import datetime
```
Magus6784, line 1: No need import datetime here
And i think we need:
from app.cafe import Cafe

Magus6784 (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy:
+        return f"Friends should buy {masks_to_buy} masks"
+    else:
```
rotsen18, line 18: You dont need else statement

rotsen18 (approved):

app/main.py
TPolina: agree

TPolina (approved):

PR #38, opened 2022-05-28
Solution

4ndyua:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+from app.errors import NotVaccinatedError, \
+    OutdatedVaccineError, NotWearingMaskError
```
Flashmobber, line 3: it is better to take all classes in parentheses
so you can specify each class from the next line and not use "\" to move to the next line
and it will look better

Flashmobber (changes requested):

Flashmobber (approved):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if not("vaccine" in visitor):
+            raise \
+                NotVaccinatedError(f"{visitor['name']} is not vaccinated")
```
TPolina, line 14: better not transfer lines with `\`
better follow this style:
```
raise NotVaccinatedError(
    f"{visitor['name']} is not vaccinated"
)
```
Fix this throughout your code

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+        if visitor["vaccine"]['expiration_date'] < datetime.date.today():
```
TPolina, line 15: follow one quotes style

TPolina (changes requested):

TPolina (approved):

PR #42, opened 2022-05-29
solution

BohdanKryven:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
+import datetime
```
Hectorovich, line 1: In this task you use only "date" from "datetime" library, so you can use this construction: "from datetime import date". And use it like this: "date.today()"

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+class Cafe:
+    def __init__(self, name):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+
+        if (
+                datetime.date.today() - visitor["vaccine"]["expiration_date"]
```
Hectorovich, line 17: Here you don`t need such complex construction. You can just compare dates like this: "visitor["vaccine"]["expiration_date"] < date.today()".

Hectorovich (changes requested):

app/cafe.py
BohdanKryven: Thanks! Fixed all troubles, it's really more simple and readable

app/cafe.py
```diff
@@ -0,0 +1,22 @@
+from datetime import date
+
+from app.errors import NotVaccinatedError, \
+    OutdatedVaccineError, \
+    NotWearingMaskError
```
rotsen18, line 5: better way is to use brackets

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Cafe:
+    def __init__(self, name):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
```
rotsen18, line 14: add descriptive error message

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+        if date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError
```
rotsen18, line 17: add descriptive error message

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError
```
rotsen18, line 20: add descriptive error message

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe) -> str:
```
rotsen18, line 4: if you use type annotation add it for cafe also

rotsen18 (changes requested):

app/main.py
BohdanKryven: Thanks! Fixed

Evgeniy994 (approved):

MasterpieceElbow (approved):

PR #43, opened 2022-05-29
Katya's Solution

katevasilenko:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
```
Hectorovich, line 1: In this task you use only "date" from "datetime" library, so you can use this construction: "from datetime import date". And use it like this: "date.today()"

Hectorovich (approved):

app/cafe.py
katevasilenko: thank you for reasonable advice

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+from app.errors import \
+    NotVaccinatedError, \
+    OutdatedVaccineError, \
+    NotWearingMaskError
```
TPolina, line 5: better list imports in brackets:
```
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Cafe:
+
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
```
TPolina, line 15: add descriptive error message

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError
```
TPolina, line 17: add a descriptive error message

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError
```
TPolina, line 19: add a descriptive error message

TPolina (changes requested):

TPolina (approved):

PR #47, opened 2022-05-30
'lockdown-cafe'

CheshkoSergey:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Cafe:
+
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Not vaccinated")
```
RomaTsybulia, line 14: In this function you can return name verified visitor

TPolina (approved):
Nice!

PR #50, opened 2022-05-30
Solution

Chardona:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from datetime import date
+
+from app.errors import NotVaccinatedError, \
+    OutdatedVaccineError, NotWearingMaskError
```
rotsen18, line 4: better way is put it in brackets

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Cafe:
+    def __init__(self,
+                 name: str):
```
rotsen18, line 9: I would left args in one line

rotsen18 (changes requested):

app/cafe.py
Chardona: possible, but not obligated way.

app/cafe.py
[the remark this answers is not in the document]
Chardona: matter of personal code style. I prefer writing constructors in this manner.

Chardona: matter of personal code style. I prefer writing constructors in this manner.

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe):
+    lacking_masks = 0
+    allow = True
```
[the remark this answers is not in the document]
Chardona: Ouch, hurts. Well. That's my implementation, so can't help but disagree. I've used writing in this manner, so see nothing horrible about it.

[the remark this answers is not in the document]
Chardona: Ouch, hurts. Well. That's my implementation, so can't help but disagree. I've used writing in this manner, so see nothing horrible about it.

rotsen18 (approved):
I'm not trying to hurt you, I'm just explaining what good code style looks like. Whether or not to follow it is your choice. 
Let's see what the mentors will say. @TPolina @MasterpieceElbow

app/main.py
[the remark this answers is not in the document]
TPolina: flags can be necessary in some cases but here it's redundant. When `NotWearingMaskError` occurs just increment `lacking_masks`. Then check if there is at least one `lacking_masks` and return `Friends should buy masks`

app/main.py
```diff
@@ -1 +1,20 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            allow = False
+            if not friend["wearing_a_mask"]:
```
TPolina, line 15: this check is redundant as well

app/main.py
```diff
@@ -1 +1,20 @@
...
+                lacking_masks += 1
+    if allow:
+        return f"Friends can go to {cafe.name}"
+    else:
```
TPolina, line 19: with `return` you don't need `else`

TPolina (changes requested):

MasterpieceElbow (approved):

PR #53, opened 2022-05-30
Solution

Kiawami:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+class Cafe:
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(f"{visitor['name']} is not vaccinated")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError(
+                f"{visitor['name']}'s vaccine is expired")
```
rotsen18, line 16: I'm not sure but this way maybe better
```suggestion
                f'{visitor["name"]}\'s vaccine is expired')

```

rotsen18 (approved):

app/cafe.py
TPolina: both variants are ok. I prefer the first one

TPolina (approved):
Well done!

PR #54, opened 2022-05-30
Solution

Evgeniy994:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+class Cafe:
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError
```
rotsen18, line 18: add descriptive error message

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+        elif visitor["wearing_a_mask"] is False:
```
rotsen18, line 19: it's more pythonic way ;)
```suggestion
        elif not visitor["wearing_a_mask"]:

```

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy != 0:
```
rotsen18, line 16: just write
```suggestion
    if not masks_to_buy:
```

rotsen18 (changes requested):

app/main.py
Evgeniy994: Done

app/main.py
```diff
@@ -13,7 +13,7 @@ def go_to_cafe(friends: list, cafe: Cafe):
         except NotWearingMaskError:
             masks_to_buy += 1
 
-    if not masks_to_buy:
+    if masks_to_buy > 0:
```
rotsen18: Yes, you are wright, this should be "if masks_to_buy:" - without not. Also you don't need the last else statement. You can write just return f"Friends can go to {cafe.name}".
Good job!

TPolina (approved):
Nice!

PR #66, opened 2022-06-20
Solution

artemkazakov947:
Thank for help

### DIFF
--- a/app/cafe.py
+++ b/app/cafe.py
@@ -1,0 +1,17 @@
+import datetime
+from app.errors import NotVaccinatedError, OutdatedVaccineError
+from app.errors import NotWearingMaskError
+
+
+class Cafe:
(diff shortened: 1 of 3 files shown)


app/cafe.py
```diff
@@ -0,0 +1,17 @@
+import datetime
+from app.errors import NotVaccinatedError, OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
IvanRamyk, line 3: Use one `import` to import all errors. Use such style if there are a lot of classes
```
from app.errors import (
     NotVaccinatedError,
     OutdatedVaccineError,
     NotWearingMaskError
)
```

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #78, opened 2022-07-21
Solution

Vlad-Peresta:


app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+        if "vaccine" not in visitor:
+            raise errors.NotVaccinatedError("Person is not vaccinated")
+
+        elif today > visitor["vaccine"]["expiration_date"]:
+            raise errors.OutdatedVaccineError("The vaccine had expired")
+
+        elif not visitor["wearing_a_mask"]:
+            raise errors.NotWearingMaskError("The visitor without mask")
+
+        else:
```
IvanRamyk, line 22: else is redundant here

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+    def visit_cafe(self, visitor: dict):
+        today = datetime.date.today()
+
+        if "vaccine" not in visitor:
+            raise errors.NotVaccinatedError("Person is not vaccinated")
+
+        elif today > visitor["vaccine"]["expiration_date"]:
+            raise errors.OutdatedVaccineError("The vaccine had expired")
+
+        elif not visitor["wearing_a_mask"]:
```
IvanRamyk, line 19: Use `if` instead of `elif`

You can do so because if the previous condition is met the previous error is already raised

IvanRamyk (changes requested):

app/cafe.py
Vlad-Peresta: thank you very much)!

IvanRamyk (approved):
Well done!

app/cafe.py
IvanRamyk: Welcome :)

Abnormaltype (approved):

PR #91, opened 2022-08-18
'Solution'

Diliaver:


app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import datetime
+from app.errors import NotVaccinatedError, \
+    OutdatedVaccineError, NotWearingMaskError
```
DmytroSvirsa, line 3: to make it look better you can use `()`

DmytroSvirsa (approved):

PR #93, opened 2022-08-20
My Solution

YanPurdenko:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+class Cafe:
+
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        today = datetime.date.today()
+        visitor_keys = [i for i in visitor.keys()]
+        if "vaccine" not in visitor_keys:
+            raise NotVaccinatedError
```
Abnormaltype, line 16: add a message about this error

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+        if visitor["vaccine"]["expiration_date"] < today:
+            raise OutdatedVaccineError
```
Abnormaltype, line 18: add a message about this error

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+        if visitor["wearing_a_mask"] is False:
+            raise NotWearingMaskError
```
Abnormaltype, line 20: add a message about this error

Abnormaltype (changes requested):

app/cafe.py
YanPurdenko: done

Abnormaltype (changes requested):
all tests should are pass

Abnormaltype (approved):
Well done!

PR #94, opened 2022-08-21
Solution

volodymyr-vereshchak:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+    OutdatedVaccineError, NotWearingMaskError
+
+
+class Cafe:
+    def __init__(self, name: str):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if 'vaccine' not in visitor:
+            raise NotVaccinatedError
```
Abnormaltype, line 12: you should add a message about this error

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+        if visitor['vaccine']['expiration_date'] < date.today():
+            raise OutdatedVaccineError
```
Abnormaltype, line 14: you should add a message about this error

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+        if 'wearing_a_mask' not in visitor or\
+                visitor['wearing_a_mask'] is False:
+            raise NotWearingMaskError
```
Abnormaltype, line 17: you should add a message about this error

Abnormaltype (changes requested):
use `""` instead of `''`

PR #109, opened 2022-09-17
Solution

hai-le-di:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict):
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError
+        else:
```
IvanRamyk, line 19: Please, remove the redundant else

app/main.py
```diff
@@ -1 +1,19 @@
...
+        try:
+            cafe.visit_cafe(friend)
+            count += 1
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy > 0:
+        return f"Friends should buy {masks_to_buy} masks"
+    if count == len(friends):
```
IvanRamyk, line 18: I believe that `count` variable and this condition are extra.
If it was at least one exception then you will already return something

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #120, opened 2022-09-23
solution

dkyrpan:


app/main.py
```diff
@@ -1 +1,17 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy > 0:
+        return f"Friends should buy {masks_to_buy} masks"
+    else:
+        return f"Friends can go to {cafe.name}"
```
DmytroSvirsa, line 17: you can avoid else with return statement

DmytroSvirsa (changes requested):

app/main.py
dkyrpan: ok

DmytroSvirsa (approved):

PR #121, opened 2022-09-24
'Solution'

AVyha:


mltdll (approved):

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        try:
+            expiration_date = visitor["vaccine"]["expiration_date"]
+        except Exception:
```
Nattalli, line 14: better to do:
```python
if "vaccine" not in visitor:
```

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+            raise NotVaccinatedError
```
Nattalli, line 15: add a message about this an error

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+        if datetime.date.today() > expiration_date:
+            raise OutdatedVaccineError
```
Nattalli, line 17: add a message about this an error

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError
```
Nattalli, line 19: add a message about this an error

Nattalli (changes requested):

Nattalli (approved):
Excellent! 🌞

PR #144, opened 2022-09-30
Solution

KatrinLazarenko:


Kitazuka (approved):
Looks good

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> [NotVaccinatedError,
+                                            OutdatedVaccineError,
+                                            NotWearingMaskError,
+                                            str]:
```
TPolina, line 14: `raise` and `return` are not the same. You don't return these errors

app/main.py
```diff
@@ -1 +1,15 @@
...
+    missing_masks = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            missing_masks += 1
+    return (f"Friends should buy {missing_masks} masks" if missing_masks > 0
+            else f"Friends can go to {cafe.name}")
```
TPolina, line 15: better use simple `if` here because it will be more readable

TPolina (changes requested):

TPolina (approved):
Good job!

PR #145, opened 2022-09-30
Solution

Kitazuka:


KatrinLazarenko (approved):
For me looks good! (except warnings in fun visit_ cafe but it is a matter of taste)

PR #146, opened 2022-09-30
Solution

Mowka7:


mltdll (approved):

app/cafe.py
```diff
@@ -0,0 +1,18 @@
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
TPolina, line 3: code duplication, better mention errors in brackets:
```
from app.errors import (
    ... ,
    ... ,
    ...
)
```

app/errors.py
```diff
@@ -0,0 +1,16 @@
+class VaccineError(Exception):
+    def __str__(self) -> str:
+        return "All friends should be vaccinated"
```
TPolina, line 3: it's not ok to have one error message for two different errors. Also, `NotVaccinatedError` and `OutdatedVaccineError` are called for one person and it's not logical to write "all friends"

app/errors.py
```diff
@@ -0,0 +1,16 @@
...
+    pass
+
+
+class OutdatedVaccineError(VaccineError):
+    pass
+
+
+class NotWearingMaskError(Exception):
+    def __str__(self) -> str:
+        return "All friends should wear the mask"
```
TPolina, line 16: same here: why do "all friends" when the error is called for one person?

app/main.py
```diff
@@ -1 +1,18 @@
...
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+
+    for person in friends:
+        try:
+            cafe.visit_cafe(person)
+        except VaccineError as e:
+            return str(e)
```
TPolina, line 12: and here you can finally write "all friends" because this function works for a list of friends

TPolina (changes requested):

TPolina (approved):
Well done!

PR #150, opened 2022-09-30
Solved the task

Cerne13:


mltdll (approved):

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name):
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if 'vaccine' not in visitor:
```
TPolina, line 15: please, use one quote style in your code (double quotes are preferable)

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name):
```
TPolina, line 11: please, add type annotation where it is missing

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if 'vaccine' not in visitor:
+            raise NotVaccinatedError
+
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
TPolina, line 18: it's better to use `if` and not `elif` because logically few errors can be present at the same time and the interpreter should be able to access all three blocks of code.

app/errors.py
```diff
@@ -0,0 +1,17 @@
+class VaccineError(Exception):
+    pass
+
+
+class NotVaccinatedError(VaccineError):
+    def __str__(self):
+        return "All friends should be vaccinated"
```
TPolina, line 7: this error is called for one person and it's not logical to have message that all friends should be vaccinated

TPolina (changes requested):

app/errors.py
Cerne13: Thank you for all the remarks!
(I guess here I just copypasted the string from readme.md)

TPolina (approved):
Good job!

PR #152, opened 2022-09-30
'Solution'

Ilyakson:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.errors import (VaccineError, NotWearingMaskError)
```
mltdll, line 1: Parentheses are unnecessary here, since you don't have to break lines.

mltdll (approved):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("The guest without vaccine")
+        elif visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("The vaccine is outdated")
```
TPolina, line 18: it's better to use `if` here instead of `elif` because logically few errors can happen at the same time and interpreter should access all blocks of code

TPolina (changes requested):

TPolina (approved):
Well done!

PR #154, opened 2022-09-30
Created composition of classes with error handlings

alexzubkoff:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
+import datetime
+from app.errors import NotVaccinatedError, \
```
mltdll, line 2: You should use `()` instead of `\`:
```python
from app.errors import (NotVaccinatedError,
                        OutdatedVaccineError,
                        NotWearingMaskError)
```
or
```python
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)

File: app/cafe.py [hunk 2]
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("All friends should be vaccinated")
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("NotWearingMaskError")
```
mltdll, line 17: This error message is not in same format as the other two. Consider changing it.

app/errors.py
```diff
@@ -0,0 +1,18 @@
+class Error(Exception):
```
mltdll, line 1: This class is unnecessary. Remove it and inherit `VaccineError` and `NotWearingMaskError` directly from `Exception`.

app/main.py
```diff
@@ -1 +1,21 @@
...
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    masks_to_buy = 0
+    try:
+        for friend in friends:
+            try:
+                cafe.visit_cafe(friend)
+            except NotWearingMaskError:
+                masks_to_buy += 1
+            except VaccineError:
+                raise
```
mltdll, line 14: You don't need to re-raise exception here. Since your `VaccineError` instances have message you have to return, you can do this:
```python
except VaccineError as e:
    return str(e)
```
and get rid of the outer `try ... except` block.

app/main.py
```diff
@@ -1 +1,21 @@
...
+    except VaccineError as e:
+        message = ''.join(list(e.args))
+        return message
+    else:
```
mltdll, line 18: There's no point in using `else` condition after `return` in preceding `if`statement. You can do it like this:
```python
if ...:
    return ...

return ...

**Review by mltdll — CHANGES_REQUESTED** [review]

File: app/main.py (hunk 5 above)
Comment (alexzubkoff) [reply app/main.py]: Thanks, for the good advice!

File: app/cafe.py (hunk 1 above)
Comment (alexzubkoff) [reply app/cafe.py]: Thanks, for the good advice!

File: app/main.py [hunk 6]
```diff
@@ -4,18 +4,14 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except NotWearingMaskError:
+            masks_to_buy += 1
+        except VaccineError as e:
+            return str(e)
+
+    if masks_to_buy != 0:
+        return f"Friends should buy {masks_to_buy} masks"
+    return "Friends can go to KFC"
```
mltdll, line 17: Also, here function should return "Friends can go to {cafe.name}", not hard-coded KFC.

mltdll (changes requested):

app/main.py
alexzubkoff: Thanks, for the good advice!

mltdll (approved):

app/main.py
alexzubkoff: Agree with you. Thanks for your vigilance!

app/main.py
```diff
@@ -1 +1,17 @@
...
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotWearingMaskError:
+            masks_to_buy += 1
+        except VaccineError as e:
+            return str(e)
+
+    if masks_to_buy != 0:
```
Y-Havryliv, line 15: `!= 0` is redundant here, use `if masks_to_buy:`

Y-Havryliv (changes requested):

alexzubkoff:
Ок! Thanks!

Пн, 3 окт. 2022 г. в 11:01, Yaroslav ***@***.***>:

Y-Havryliv (approved):
GJ!

PR #155, opened 2022-09-30
Solution

Qusert17:


mltdll (approved):

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Go get a Vaccine")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("Vaccine is not valid go get a new one")
+        if visitor["wearing_a_mask"] is False:
+            raise NotWearingMaskError("Feel free to buy a mask")
+        else:
+            return f"Welcome to {self.name}"
```
TPolina, line 20: `else` is unnecessary here. You can just use the `return` statement because if an error happens the function will end and only if no errors return will be executed

TPolina (changes requested):

app/cafe.py
Qusert17: Done)

TPolina (approved):
Well done!

PR #158, opened 2022-10-01
Solution

VadymShkarbul:


app/errors.py
```diff
@@ -0,0 +1,17 @@
+class VaccineError(Exception):
+    pass
+
+
+class NotVaccinatedError(VaccineError):
+    def __str__(self):
```
Vasyl-Poremchuk, line 6: I think this magic method is not needed here, you just need to write `pass` operator as you write in the previous exception or use the `docstring`. Check it in all class exceptions of this module

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from app.cafe import Cafe
+# from app.errors import * bad practice or something?
```
Vasyl-Poremchuk, line 2: Maybe you should delete this comment

app/main.py
```diff
@@ -1 +1,19 @@
...
+from app.errors import NotVaccinatedError, OutdatedVaccineError
```
Vasyl-Poremchuk, line 3: You do not need to import both of these classes, because they are subclasses of the `VaccineError` superclass. You can write: `from app.errors import VaccineError, NotWearingMaskError` and delete the next line

app/main.py
```diff
@@ -1 +1,19 @@
...
+    for visitor in friends:
+        try:
+            cafe.visit_cafe(visitor)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            without_mask += 1
+    if without_mask > 0:
+        return f"Friends should buy {without_mask} masks"
+    else:
```
Vasyl-Poremchuk, line 18: And do not use `else` condition because if the previous condition is not True you will return this expression anyway, just write `return f"Friends can go to {cafe.name}"` and shift it to the left by 4 spaces

Vasyl-Poremchuk (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,18 @@
+from app.errors import NotVaccinatedError, OutdatedVaccineError
+from app.errors import NotWearingMaskError
+import datetime
```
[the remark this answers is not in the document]
VadymShkarbul: Thank you for review. Maybe I missed something about module order in general

app/main.py
VadymShkarbul: No it was a question for mentors, I wanted to add all from errors this way but flake8 didn't agree with me.

app/main.py
VadymShkarbul: Yes thank you!

app/errors.py
VadymShkarbul: I think that this is just one of the solutions. I did this on purpose in the exceptions module to remove needless from the class method. But for example, if you suddenly need to use these exceptions again - it would not be necessary to rewrite all strings.

app/main.py
VadymShkarbul: You're absolutely right. Thank you!

app/errors.py
Vasyl-Poremchuk: Maybe you're right

Vasyl-Poremchuk (approved):

Y-Havryliv (changes requested):

app/main.py
VadymShkarbul: Thank you!

Y-Havryliv (approved):
Well done!

PR #161, opened 2022-10-01
Solution

misha-doroshenko:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+from datetime import date
+from app.errors import \
```
Ilyakson, line 2: Use Parentheses

Ilyakson (changes requested):

app/cafe.py
misha-doroshenko: Can you approve my solution?

mltdll (approved):

Ilyakson (approved):

PR #162, opened 2022-10-01
Solution

VitaliiPiaskivskyi:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
```
Vasyl-Poremchuk, line 1: Add a blank line between the different import groups
https://peps.python.org/pep-0008/#blank-lines

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+from app.errors import NotVaccinatedError, OutdatedVaccineError
```
Vasyl-Poremchuk, line 2: You can import classes from the same module like this (check your `checklist` file in this task):
```
from app.errors import (
   NotVaccinatedError,
   OutdatedVaccineError,
   NotWearingMaskError
)
```

app/main.py
```diff
@@ -1 +1,16 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    you_need_mask = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            you_need_mask += 1
+    if you_need_mask > 0:
```
Vasyl-Poremchuk, line 14: You can use `if you_need_mask:`

Vasyl-Poremchuk (changes requested):

app/cafe.py
VitaliiPiaskivskyi: ok

app/cafe.py
VitaliiPiaskivskyi: ok

Vasyl-Poremchuk (approved):
Well done!

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
```
Y-Havryliv, line 13: you don't need `.keys()` here

app/main.py
Y-Havryliv: I agree with Vasyl, `> 0` is redundant and doesn't affect understanding

Y-Havryliv (changes requested):

PR #165, opened 2022-10-02
Solution

dmytro-petliovanyi:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (VaccineError, NotWearingMaskError)
```
Vasyl-Poremchuk, line 2: Brackets are required if the string contains more than 79 characters and you want to format it according to the guidelines PEP8

Vasyl-Poremchuk (changes requested):

Vasyl-Poremchuk (approved):
Well done!

Y-Havryliv (approved):
Well done!

PR #167, opened 2022-10-02
Solution

YevheniiDavydenko:


app/main.py
```diff
@@ -1 +1,20 @@
...
+
+def go_to_cafe(friends: list, cafe: str):
+    selected_cafe = Cafe(cafe)
+    some_arr = []
+    masks_to_buy = 0
+    for person in friends:
+        try:
+            selected_cafe.visit_cafe(person)
+            some_arr.append(person)
+        except VaccineError:
```
Cerne13, line 13: Maybe separate messages for outdated vaccine and no vaccine at all should be more informative, but the solution is eloquent and according to terms. 😃

Cerne13 (approved):

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: str):
```
TPolina, line 5: according to the task description cafe is a Cafe instance

app/main.py
```diff
@@ -1 +1,20 @@
...
+    selected_cafe = Cafe(cafe)
+    some_arr = []
+    masks_to_buy = 0
+    for person in friends:
+        try:
+            selected_cafe.visit_cafe(person)
+            some_arr.append(person)
```
TPolina, line 12: there is no need to add person to the list and do some checking. Because if no Vaccine errors happen it already means that friends can go to cafe

TPolina (changes requested):

TPolina (approved):
Well done!

PR #169, opened 2022-10-03
'Solution'

illiariaboy94:


VadymShkarbul (approved):
Nice!

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Guest is not vaccinated")
+        elif datetime.date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("Vaccine is outdated")
+        elif not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("Guest without mask")
+        else:
```
Y-Havryliv, line 18: `else` is redundant and you can use `if` instead of `elif`

Y-Havryliv (changes requested):

app/cafe.py
illiariaboy94: Ok, thanks

Y-Havryliv (approved):
Nice!

PR #173, opened 2022-10-03
Solution

LiudmylaKulinchenko:


app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
mltdll, line 19: You can just use `if masks_to_buy` in this case.

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (
```
mltdll, line 2: The imports can probably fit into one line.

mltdll (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitors must have a vaccine!")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
mltdll, line 17: Also, you don't have to use elif here. Don't use `elif` or `else` after you `raise` an exception or `return` in an if block.

app/main.py
LiudmylaKulinchenko: I like how it looks and don't think this is an error

mltdll (approved):
GJ!

TPolina (approved):
Well done!

PR #175, opened 2022-10-06
'Solution'

IvanGLS:


app/main.py
```diff
@@ -1 +1,17 @@
...
+        try:
+            cafe.visit_cafe(visitor)
+        except app.errors.VaccineError:
+            return "All friends should be vaccinated"
+        except app.errors.NotWearingMaskError as e:
+            masks_to_buy += 1
+            error_message = e
+    if masks_to_buy >= 1:
+        return f"{error_message} {masks_to_buy} masks"
+    else:
```
Abnormaltype, line 16: remove unnecessary else

app/cafe.py
```diff
@@ -0,0 +1,17 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise error.NotVaccinatedError
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise error.OutdatedVaccineError
+        if visitor["wearing_a_mask"] is False:
+            raise error.NotWearingMaskError
+        else:
```
Abnormaltype, line 16: remove unnecessary else

app/errors.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class NotVaccinatedError(VaccineError):
+    """NotVaccinatedError"""
+
+
+class OutdatedVaccineError(VaccineError):
+    """OutdatedVaccineError"""
+
+
```
Abnormaltype, line 13: add a  message to these errors

Abnormaltype (changes requested):

app/cafe.py
IvanGLS: How can I add that without else?
Screenshot (85)

app/errors.py
```diff
@@ -0,0 +1,18 @@
...
+class NotVaccinatedError(VaccineError):
+    def __str__(self) -> str:
+        return "All friends should be vaccinated"
+
+
+class OutdatedVaccineError(VaccineError):
+    def __str__(self) -> str:
+        return "All friends should be vaccinated"
```
Abnormaltype, line 13: you have three different Errors and one message for them. And the logic question, do you need 3 Errors?)

Abnormaltype (changes requested):

app/errors.py
IvanGLS: How it's can be? If message the same from task.
Screenshot (87)
Screenshot (86)

Abnormaltype: yes, but create different messages for these errors

Abnormaltype (approved):
Well done!

PR #179, opened 2022-10-11
Solution of py-lockdown-cafe

YLobunets:


app/main.py
```diff
@@ -1 +1,22 @@
...
+            _ = cafe.visit_cafe(friend)
+            friends_to_ok += 1
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if friends_to_ok == len(friends):
+        return f"Friends can go to {cafe.name}"
+    else:
```
IvanRamyk, line 21: Please, remove the redundant else

IvanRamyk (changes requested):

app/main.py
YLobunets: Сonstruction "else" was removed

app/main.py
```diff
@@ -1 +1,21 @@
...
+from .cafe import Cafe
+from .errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    friends_to_ok = 0
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            _ = cafe.visit_cafe(friend)
```
IvanRamyk, line 12: `_` - unclear variable name :)

IvanRamyk (changes requested):

app/main.py
YLobunets: Extra variable removed

PR #188, opened 2022-10-29
Cafe

bythewaters:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+import datetime
+from app.errors \
```
IvanRamyk, line 2: use brackets instead of backslash due to PEP8 recommendations

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+class Cafe:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError")
+
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
IvanRamyk, line 15:
```suggestion
        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```

app/main.py
```diff
@@ -1 +1,25 @@
...
+        except NotWearingMaskError:
+            count_mask += 1
+
+        else:
+            count += 1
+
+    if count_mask != 0:
+        return f"Friends should buy {count_mask} masks"
+
+    if count == 2:
```
IvanRamyk, line 24: what if there are more friends?

IvanRamyk (changes requested):

app/cafe.py
```diff
@@ -1,6 +1,7 @@
 import datetime
-from app.errors \
-    import OutdatedVaccineError, NotWearingMaskError, NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
+from app.errors import NotVaccinatedError
```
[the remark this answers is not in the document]
IvanRamyk: image
Sorry, it wasn't in pep8 but in https://peps.python.org/pep-0328/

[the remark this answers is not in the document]
IvanRamyk: Use standard group mechanism

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #195, opened 2022-11-09
Solution

Ida83:


app/cafe.py
```diff
@@ -0,0 +1,17 @@
+import datetime
+from app.errors import NotVaccinatedError, \
+    OutdatedVaccineError, NotWearingMaskError
```
DmytroSvirsa, line 3: use () for better code style of import

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):
GJ!

PR #205, opened 2022-11-23
Solution

ArtMot99:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+from app.errors import NotVaccinatedError, \
+    OutdatedVaccineError, \
+    NotWearingMaskError
```
Abnormaltype, line 4: use checlist

Abnormaltype (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
OleksandrPozden, line 14: it is better to use if instead of elif

app/main.py
```diff
@@ -1 +1,20 @@
...
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(visitor=friend)
+
```
OleksandrPozden, line 10: remove redundant space between the lines

OleksandrPozden (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks = 0
```
IvanRamyk, line 6:
```suggestion
    masks_to_buy = 0
```

IvanRamyk: just a bit clear name

IvanRamyk (changes requested):

IvanRamyk (approved):
but it's already ok 
Good job!

PR #208, opened 2022-11-26
Solution

ViktorKrUA:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError
+                        )
```
YuriiKindrat, line 5: fix the import according to the best practicies

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("A visitor should be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("A visitor should be vaccinated")
```
YuriiKindrat, line 16: write more informative message for this error, f.e. "A visitor should have up-to-date vaccine"

YuriiKindrat (changes requested):

app/cafe.py
ViktorKrUA: here is in app.errors module we don't have too many classes to be imported... so, I would like to use * 
from app.errors import *

please let me know if in this case it will be exactable

app/cafe.py
ViktorKrUA: thanks for this feedback..  It will be revised

app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
YuriiKindrat, line 4: U can use here only one import here, but u should use best practices for breaking a line in import

YuriiKindrat (changes requested):

app/cafe.py
YuriiKindrat: It is better not to use * in imports

app/cafe.py
YuriiKindrat: from package_name import (
    x,
    y,
    z,
)

ViktorKrUA: Much obliged!

YuriiKindrat (approved):

PR #214, opened 2022-12-03
Solution

viktory-koroliova:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise errors.NotVaccinatedError(
+                "Everyone should be vaccinated")
```
Serhii-Leonenko, line 12:
```suggestion
            raise errors.NotVaccinatedError(
                "Everyone should be vaccinated"
                )
```

The same in other similar cases

app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+from app import errors
```
Serhii-Leonenko, line 2: Use multiple imports:
`from app.errors import (NotVaccinatedError...)`

Serhii-Leonenko: Like an example in your checklist 
image

app/main.py
```diff
@@ -1 +1,20 @@
...
+        except errors.VaccineError:
+            count_vaccine += 1
+        except errors.NotWearingMaskError:
+            count_mask += 1
+    if count_mask == 0 and count_vaccine == 0:
+        return f"Friends can go to {cafe.name}"
+    if count_vaccine > 0:
+        return "All friends should be vaccinated"
+    if count_vaccine == 0 and count_mask > 0:
+        return f"Friends should buy {count_mask} masks"
```
Serhii-Leonenko, line 20: `count_vaccine = 0` is unnecessary
When you excepting `VaccineError` just return a message  "All friends should be vaccinated"
So then you can remove conditions `if count_vaccine > 0:` and `if count_mask == 0 and count_vaccine == 0:`

Serhii-Leonenko (changes requested):

OleksandrPozden (approved):
Good!

PR #220, opened 2022-12-13
solution

kleotan901:


app/main.py
```diff
@@ -1 +1,27 @@
...
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    count_vaccinated_error = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
YuriiKindrat, line 18: image
fix this according to the task

YuriiKindrat (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    name = ""
```
IvanRamyk, line 11: why the name is a **class** attribute?

app/main.py
```diff
@@ -1 +1,26 @@
...
+    count_vaccinated_error = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            count_vaccinated_error += 1
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if count_vaccinated_error:
+        return "All friends should be vaccinated"
```
IvanRamyk, line 22: you can return this message already in line 18

IvanRamyk: if smbd isn't vaccinated there is no need to continue iterating

YuriiKindrat (approved):

PR #221, opened 2022-12-16
Solution

yuraua5:


app/main.py
```diff
@@ -1 +1,19 @@
...
+            cafe.visit_cafe(friend)
+            correct_count += 1
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            not_wearing_mask += 1
+    if not_wearing_mask:
+        return f"Friends should buy {not_wearing_mask} masks"
+    if correct_count == len(friends):
+        return f"Friends can go to {cafe.name}"
```
YuriiKindrat, line 19: No need to make correct_count variable, u can just left a return with "friends can go to cafe" without if condition

YuriiKindrat (changes requested):

OleksandrPozden (approved):

PR #224, opened 2022-12-19
Develop

sorrow03:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from datetime import date
+
+
+from app.errors import NotVaccinatedError, \
+    NotWearingMaskError, OutdatedVaccineError
```
katevasilenko, line 5: It’s better to use brackets
`from app.errors import (
    NotVaccinatedError,
    NotWearingMaskError,
    OutdatedVaccineError
    ) `

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from datetime import date
+
+
```
katevasilenko, line 3: One empty line is enough

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+from app.cafe import Cafe
```
katevasilenko, line 2: I would prefer `app.cafe` on top as an alphabetic sorted

katevasilenko (changes requested):
Good job, I have only few picky request to change code style

katevasilenko (approved):
Good job!

PR #227, opened 2022-12-19
Done

phaishuk:
Really don't know how to write "go to cafe" with one cycle

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,27 @@
-# write your code here
+from __future__ import annotations
+from app.cafe import Cafe
+from app.errors import (
+    VaccineError,
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str | VaccineError:
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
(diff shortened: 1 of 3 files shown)


app/main.py
```diff
@@ -1 +1,27 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotWearingMaskError:
+            unmasked = 0
+            for some_friend in friends:
+                if some_friend["wearing_a_mask"] is False:
+                    unmasked += 1
+            return f"Friends should buy {unmasked} masks"
+    return f"Friends can go to {cafe.name}"
```
Abnormaltype, line 27: why did you use two loops and try/except statements?

Abnormaltype (changes requested):

app/main.py
phaishuk: Because i have to move over all the friends in income dict. The first loop catch vaccine error, the second masks error, so because of order i've done two loops. Is it posible to avoid loops or all the loops?

Abnormaltype: yep) write in the chat in slack)

OleksandrPozden (approved):

PR #240, opened 2023-01-06
Solution

yuliia-stopkyna:


app/main.py
```diff
@@ -1 +1,21 @@
...
+    masks_to_buy = 0
+    try:
+        for friend in friends:
+            try:
+                cafe.visit_cafe(friend)
+            except NotWearingMaskError:
+                masks_to_buy += 1
+    except VaccineError:
+        return "All friends should be vaccinated"
+    if masks_to_buy:
```
Vasyl-Smutok, line 19: better to use something obvious

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(f"Only vaccinated people "
+                                     f"can visit {self.name}")
```
Vasyl-Smutok, line 17: think so, it will look better

```
raise NotVaccinatedError(
    f"Only vaccinated people can visit {self.name}"
)
```

Vasyl-Smutok (changes requested):

app/main.py
yuliia-stopkyna: @Vasyl-Smutok what do you mean? in condition?

Vasyl-Smutok: `if masks_to_buy != 0:` or `if masks_to_buy >= 1:`

Vasyl-Smutok (approved):
Well done 👍

app/main.py
yuliia-stopkyna: I followed the checklist for this task - and there is stated that it's better not to compare with 0.

PR #252, opened 2023-01-19
'Solution'

timyr4iq:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError
+        if visitor["wearing_a_mask"] is False:
```
ArsenPidhoretskyi, line 16: if wearing_a_mask does not present in visitor will raise Error

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
ArsenPidhoretskyi, line 14: if expiration_date does not present in visitor["vaccine"] will raise an Error

ArsenPidhoretskyi (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,22 @@
+import datetime
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
Vasyl-Smutok, line 4: better like this 
```
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```

app/cafe.py
Vasyl-Smutok: I think @ArsenPidhoretskyi  means that you should use "get" here

app/cafe.py
Vasyl-Smutok: I think @ArsenPidhoretskyi means that you should use "get" here

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
```
Vasyl-Smutok, line 13: added descriptions to Error

Please fix here and other places

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (VaccineError, NotWearingMaskError)
```
katevasilenko, line 2: Remove redundant brackets
```suggestion
from app.errors import VaccineError, NotWearingMaskError
```

katevasilenko (changes requested):

katevasilenko (approved):
Good job

PR #255, opened 2023-01-22
Solution

serg121192:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
+import datetime
+
+
+from app.errors import NotVaccinatedError
+from app.errors import NotWearingMaskError
+from app.errors import OutdatedVaccineError
```
olena-hul, line 6: You can use 1 import for this

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(f"{visitor['name']} must be vaccinated!")
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Vaccine should be fresh!")
```
olena-hul, line 19: Nit: fresh is related rather to food than vaccines) You can say that "The vaccine has expired" or something like this)

olena-hul (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(f"{visitor['name']} must be vaccinated!")
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Your vaccine has expired!")
+
+        if visitor["wearing_a_mask"] is False:
```
ArsenPidhoretskyi, line 23: use .get("wearing_a_mask") possible wearing_a_mask can not exist in visitor. Also you can write as:
```suggestion
        if not visitor.get("wearing_a_mask"):

```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #258, opened 2023-01-25
Solution

TarasSavchyn:


app/cafe.py
```diff
@@ -0,0 +1,27 @@
+import datetime
+
+from app.errors import NotVaccinatedError, \
+    NotWearingMaskError, OutdatedVaccineError
```
olena-hul, line 4: Use `()` for imports, not `\`

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        flag = [True]
```
olena-hul, line 12: This is not needed

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+
+        if not visitor.get("vaccine", False):
```
olena-hul, line 14:
```suggestion
        if not visitor.get("vaccine"):
```

app/main.py
```diff
@@ -1 +1,24 @@
...
+    return 0 if not friend.get("vaccine", False) or \
+        friend["vaccine"]["expiration_date"] < datetime.date.today() else 1
+
+
+def count_mask(persons: list) -> int:
+    return sum([1 for person in persons if not person["wearing_a_mask"]])
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    try:
```
olena-hul, line 16: Declare a variable here that represents masks to buy count. Then use loop to iterate through friends. Increment masks counter in except NotWearingMaskError block. The functions above are not needed, they add extra complexity.

olena-hul (changes requested):

app/main.py
Vasyl-Smutok: @TarasSavchyn 
 Please do as Olena said

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+import datetime
+
+
+def sorted_friends(friend: dict) -> int:
+    return 0 if not friend.get("vaccine", False) or \
+        friend["vaccine"]["expiration_date"] < datetime.date.today() else 1
```
Vasyl-Smutok, line 8:
```
return (
    0 if not friend.get("vaccine", False) or
    friend["vaccine"]["expiration_date"] < datetime.date.today() else 1
)
```

Vasyl-Smutok (changes requested):

app/main.py
TarasSavchyn: > Declare a variable here that represents masks to buy count. Then use loop to iterate through friends. Increment masks counter in except NotWearingMaskError block. The functions above are not needed, they add extra complexity.

I removed the count function, but without the sorting function, the logic is lost

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    vac_errors = []
+    mask_errors = []
```
Abnormaltype, line 7: unnecessary to use lists here, better to use only one variable for masks. b/c u can return in place of except when you catch VaccineError)

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #259, opened 2023-01-25
py-lockdown-cafe solution

frameq:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Cafe:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError")
```
olena-hul, line 16: Provide informative exception messages

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        if not visitor["wearing_a_mask"]:
```
olena-hul, line 19:
```suggestion
        if not visitor.get("wearing_a_mask"):
```

olena-hul (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+class Cafe:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+        if (visitor.get("vaccine").get("expiration_date")
+                < datetime.date.today()):
```
olena-hul, line 18: Vaccine is already present, if not, the first error will be risen. So you can just leave as it was before.
If the expiration date is not present, None will be returned and you will get `TypeError: '<' not supported between instances of 'NoneType' and 'datetime.datetime'`

olena-hul (changes requested):

PR #261, opened 2023-01-31
Solution

ilushasm:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+
+import app.errors as error
```
katevasilenko, line 3: Please import every single error using ()

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise error.NotVaccinatedError
```
katevasilenko, line 12: Please provide description in every `raise` block

app/errors.py
```diff
@@ -0,0 +1,17 @@
+class NotWearingMaskError(Exception):
+    def __str__(self) -> str:
+        return "No mask — no pass"
```
katevasilenko, line 3: You can use `pass` here for all errors if you provide description

katevasilenko (changes requested):

katevasilenko (approved):
Well done!

PR #264, opened 2023-02-03
Solution

PodorogaNatalia:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+import datetime
+from typing import Union
+from app.errors import NotVaccinatedError, \
```
Grayder0152, line 3: Use `()` instead `/`

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> Union[str, None]:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All visitors should be vaccinated")
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Vaccine expired")
+
+        if visitor["wearing_a_mask"] is False:
```
Grayder0152, line 18: Use `not visitor["wearing_a_mask"]` instaed of `visitor["wearing_a_mask"] is False` it is better approach

app/main.py
```diff
@@ -1 +1,23 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    vaccine_problems = 0
+    mask_problems = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError as ve:
+            print(ve)
+            vaccine_problems += 1
```
Grayder0152, line 14: image
So you can not use `vaccine_problems` and return result in this line

Grayder0152 (changes requested):

Vasyl-Smutok (approved):
Well Done 👍

PR #280, opened 2023-02-11
Solution

manjustice:


AndereLion (approved):
OK!

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("You must have a vaccine")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("You expiration date is out")
```
YuriiKindrat, line 18: it seems to me, that is better to write here: "vaccine expiration date is out" not "you" :)

YuriiKindrat (approved):

PR #281, opened 2023-02-11
1 commit

AndereLion:


SafonovVladimir (approved):
Perfect!

manjustice (approved):

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        vaccine = visitor.get("vaccine")
+        mask = visitor.get("wearing_a_mask")
+        today = datetime.date.today()
+
+        if not vaccine:
+            raise NotVaccinatedError("All friends should be vaccinated")
```
wQuelS, line 18: I guess it should say that `visitor` is not vaccinated, not all friends should have vaccine

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+
+        if today > vaccine["expiration_date"]:
+            raise OutdatedVaccineError("All friends should be vaccinated")
```
wQuelS, line 21: visitor came with outdated vaccine not without it

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+
+        if not mask:
+            raise NotWearingMaskError("All Friends should buy masks")
```
wQuelS, line 24: again, visitor should have a mask, not `all friends`

wQuelS (changes requested):

app/cafe.py
AndereLion: Fixed

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        vaccine = visitor.get("vaccine")
+        mask = visitor.get("wearing_a_mask")
+        today = datetime.date.today()
+
+        if not vaccine:
+            raise NotVaccinatedError(f"{self.name} is not vaccinated,")
```
wQuelS, line 18: so, for example, "KFC" is not vaccinated ?

wQuelS (changes requested):

app/cafe.py
AndereLion: My bad =). Fixed.

wQuelS (approved):
Good job

PR #282, opened 2023-02-12
lockdown_cafe

OmeLchDO:


app/main.py
```diff
@@ -1 +1,16 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_needed = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_needed += 1
+    if masks_needed != 0:
```
vasylhnatiuk, line 14: will be better to use:` if masks_needed:`

vasylhnatiuk (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
+import datetime
+
+from app.errors import NotVaccinatedError, \
+    OutdatedVaccineError, NotWearingMaskError
```
Vasyl-Smutok, line 4:
```suggestion
from app.errors import(
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```

Vasyl-Smutok (changes requested):

app/cafe.py
OmeLchDO: Yes, nice comment! I agree with you

Vasyl-Smutok (approved):
Well Done 👍

PR #284, opened 2023-02-13
Solution

MaybeTI:


app/main.py
```diff
@@ -1 +1,17 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            count_maks += 1
+    if count_maks:
+        return f"Friends should buy {count_maks} masks"
+    else:
+        return f"Friends can go to {cafe.name}"
```
bihunva, line 17: You may not use "else" here

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    count_maks = 0
```
bihunva, line 6: "maks"?

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+    def visit_cafe(self, visitor: dict) -> str:
+        try:
+            if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+                raise OutdatedVaccineError("You expiration date is past due")
+            if not visitor["wearing_a_mask"]:
+                raise NotWearingMaskError("Wear a mask")
+        except KeyError:
+            raise NotVaccinatedError("You are not vaccinated")
+        else:
+            return f"Welcome to {self.name}"
```
bihunva, line 20: This block can be simplified by using three "if's" and not using "try/except/else"

bihunva (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("You are not vaccinated")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("You expiration date is past due")
+        elif not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("Wear a mask")
+        else:
```
bihunva, line 18: Again, you may not use "else" here, it would be more pythonic

bihunva (approved):
GJ :)

wQuelS (approved):
GJ

PR #286, opened 2023-02-13
Solution

AlimFreight:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError()
+
+        vaccine = visitor["vaccine"]
+        expiration_date = vaccine["expiration_date"]
+        if expiration_date < datetime.date.today():
+            raise OutdatedVaccineError(expiration_date)
+
+        if not visitor.get("wearing_a_mask", False):
```
ArsenPidhoretskyi, line 22: visitor.get() can return None 
not None it is True
default parameter False is redundant here

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done!

Grayder0152 (approved):

PR #289, opened 2023-02-14
Solution

NEkropo1:


DavidGlusch (approved):

app/cafe.py
```diff
@@ -0,0 +1,29 @@
...
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None | str:
+        if "vaccine" not in visitor.keys():
```
wQuelS, line 13: `in` keyword will be looking for `vaccine` in dictionary keys by default

wQuelS (changes requested):

app/cafe.py
NEkropo1: I was sure, that this method was implemented for fastening the execution ))
Thanks )

wQuelS (approved):
Nice

PR #291, opened 2023-02-15
Solution

DavidGlusch:


app/main.py
```diff
@@ -1 +1,22 @@
...
+    vaccine_flag = False
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotWearingMaskError:
+            masks_to_buy += 1
+        except VaccineError:
+            vaccine_flag = True
+    if vaccine_flag:
+        return "All friends should be vaccinated"
```
ZabFTFT, line 19: You can return string "All friends should be vaccinated" inside except VaccineError block.

ZabFTFT (approved):

app/main.py
ZabFTFT: But overall GJ! I also liked that you grabbed name of visitor in Cafe's visit method.Error's descriptions are also very good!

YuriiKindrat: @ZabFTFT is right, move return on 19 line below the Vaccine Error, no need to create usless variable

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #297, opened 2023-02-16
Solution

SergiiMachulin:


app/main.py
```diff
@@ -1 +1,23 @@
...
+        except VaccineError:
+            access_granted = False
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            access_granted = False
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
+        return f"Friends should buy {masks_to_buy} masks"
+    if access_granted:
```
amber-marichi, line 22: this will work but tbh i'm not sure you need the access_granted variable because it just does unnecessary double-checks: if VaccineError thrown - first return works, if mask_to_buy check > 0 - second return works, if both checks are passed then why would you need one more?

amber-marichi (approved):

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+from __future__ import annotations
```
wQuelS, line 1: do you need it ?

app/main.py
wQuelS: Agree, remove redundant check and variable

wQuelS (changes requested):

wQuelS (approved):
Good

PR #298, opened 2023-02-16
'Solution'

alexlega:


app/main.py
```diff
@@ -1 +1,21 @@
...
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str | None:
+    mask_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotWearingMaskError:
+            mask_to_buy += 1
+        except (NotVaccinatedError, OutdatedVaccineError):
+            return "All friends should be vaccinated"
```
oksana-feshchenko, line 18: "Create a VaccineError parent class for NotVaccinatedError and OutdatedVaccineError errors. Use it to catch both types of errors in the same except clause. " Use this tip here

oksana-feshchenko (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
```
wQuelS, line 1: do u need it?

app/cafe.py
```diff
@@ -0,0 +1,32 @@
...
+            raise OutdatedVaccineError(
+                f"Visitor {visitor['name']} "
+                f"should visit ambulance and update his vaccine"
+            )
+        elif ("wearing_a_mask" not in visitor
+              or visitor["wearing_a_mask"] is False):
+            raise NotWearingMaskError(
+                f"Visitor {visitor['name']} should be in mask"
+            )
+        else:
```
wQuelS, line 31: `else` is not needed here

wQuelS (changes requested):

wQuelS (approved):
Good

PR #299, opened 2023-02-16
solution

amber-marichi:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+
+from app.errors import (NotVaccinatedError, OutdatedVaccineError,
```
SergiiMachulin, line 3: I think it's better to use indentation according to the checklist example

Serhii-Leonenko: agree, fix this

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        today = datetime.date.today()
+        if visitor["vaccine"]["expiration_date"] < today:
```
Serhii-Leonenko, line 15: I think you can remove this variable and compare it directly with datetime.date.today()

Serhii-Leonenko (changes requested):

wQuelS (approved):
Cool

PR #308, opened 2023-02-16
Solution

ada-krav:


ZabFTFT (approved):
Well done! Definitely, you are a creative person with this kind of error's message =))

ada-krav:
@ZabFTFT, they said it is supposed to be descriptive :)

ZabFTFT:
> @ZabFTFT, they said it is supposed to be descriptive :)

And you accept it as a personal challenge xDD

Danil228 (approved):

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
```
wQuelS, line 14: `in` keyword will "look" in dictionary keys by default

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+            raise NotVaccinatedError(
+                f"Wow, {visitor['name']} is so late (cafe.py, line 16)."
```
wQuelS, line 16: `(cafe.py, line 16)` For what purpose?:))

wQuelS (changes requested):

app/cafe.py
ada-krav: Well, a good example in the checklist has 'raise CustomError("We have an error on the line 16.")'
I had to change the number 3 times after updating the file for flake8 recommendations. It wasn't fun :sweat_smile:

wQuelS (approved):
Nice:)

PR #313, opened 2023-02-17
Solution

yura-hudzovskyi:


app/main.py
```diff
@@ -1 +1,18 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
usrofgh, line 16: according checklist you can use here just ```if masks_to_buy``` without ```> 0```. But in generally good job

usrofgh (approved):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("All friends should be vaccinated")
```
YuriiKindrat, line 18: use more descriptive messages for this exceptions - different exceptions should have different message according to the exception name

app/main.py
YuriiKindrat: image
@usrofgh is right, fix this please

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #314, opened 2023-02-17
Solution

MikhailLyvak:


ProdyRodion (approved):
Good Job!!!

app/main.py
```diff
@@ -1 +1,17 @@
...
+def go_to_cafe(cafe: Cafe, friends: list) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotWearingMaskError:
+            masks_to_buy += 1
+        except VaccineError:
+            return "All friends should be vaccinated"
+    if masks_to_buy != 0:
```

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):

app/main.py
YuriiKindrat: image
@viktoria-rybenchuk is right, fix this please

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Someone not vaccineted")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Someone`s vaccine is outdated")
```
YuriiKindrat, line 15: u can use visitor name instead of "someone's"

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+        if visitor.get("wearing_a_mask") is False:
+            raise NotWearingMaskError("Someone haven`t mask -_0")
```
YuriiKindrat, line 17: the same here)

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #318, opened 2023-02-17
Solution

Yevheniy-Lototskiy:


app/main.py
```diff
@@ -1 +1,20 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotWearingMaskError:
+            masks_to_buy += 1
+        except VaccineError as e:
+            result = str(e)
+
+    if masks_to_buy and not result:
+        result = f"Friends should buy {masks_to_buy} masks"
```
ostapT, line 18: I don't quite understand why you need a variable result
in the 17 line you can check if mask_to_buy > 0 : return f"string"
and final return only f"string" without OR

ostapT (changes requested):

app/errors.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class NotVaccinatedError(VaccineError):
+    def __str__(self) -> str:
+        return "All friends should be vaccinated"
+
+
+class OutdatedVaccineError(VaccineError):
+    def __str__(self) -> str:
+        return "All friends should be vaccinated"
```
YuriiKindrat, line 17: image
it is not descriptive messages for exceptions, fix this, please

YuriiKindrat (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("All friends should be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("All friends should be vaccinated")
```
YuriiKindrat, line 17: it seems to me, that massages should be different for different exceptions)

Yevheniy-Lototskiy: I also wanted to do this, but the tests expect the same message for both of these cases. It should not be?

Yevheniy-Lototskiy: image

YuriiKindrat (changes requested):
fix flake8 errors

Serhii-Leonenko (approved):
Well done!

PR #320, opened 2023-02-17
Solution

Anatolii-Poznyak:


oksana0410 (approved):

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (
+    NotWearingMaskError,
+    VaccineError
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> [str, Exception]:
```
YuriiKindrat, line 8: are u sure about Exception in return ?

YuriiKindrat (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> [str, Exception]:
```
Serhii-Leonenko, line 13: Here you should fix the annotation as well. Because the function should return only string, exceptions will be raised but don't return

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Well done!

PR #328, opened 2023-02-18
Develop

AlessaQ:


oigod (approved):

app/cafe.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import annotations
```
Y-Havryliv, line 1: do you need this import?

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+from app.errors import (
```
Y-Havryliv, line 2: also the order of your imports are wrong, optimize it

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
```
Y-Havryliv, line 7: I would add some line breaks before and after for-loop to make your code more readable

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+import datetime
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | None:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError(f"{self.name} will die from corona!")
```
Y-Havryliv, line 16: RIP(

Y-Havryliv (changes requested):

YuriiKindrat (approved):

PR #330, opened 2023-02-18
'Solution'

NazarHladaniuk:


app/main.py
```diff
@@ -1 +1,16 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for visitor in friends:
+        try:
+            cafe.visit_cafe(visitor)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy != 0:
```
ramsty, line 14: you can remove "!= 0", it will be the same logic

app/cafe.py
```diff
@@ -0,0 +1,20 @@
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+import datetime
```
ramsty, line 6: The best practice is to group imports in the following order:
standard library imports
related third party imports
local application/library specific imports
So you can just change the order of imports.

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(str(NotVaccinatedError))
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError(str(OutdatedVaccineError))
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError(str(NotWearingMaskError))
```
ramsty, line 19: I think you should add a description on this exceptions

ramsty (changes requested):

ramsty (approved):

app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
```
Y-Havryliv, line 1: nice code, but two changes are needed

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for visitor in friends:
```
Y-Havryliv, line 7: I would add some line breaks before and after the for-loop to make your code more readable

Y-Havryliv (changes requested):

YuriiKindrat (approved):

PR #331, opened 2023-02-18
'Solution'

AnnaBoichenko:


me1nyk (approved):
good job, but for what you need "errors_count" in "go_to_cafe" function?
you are not using this variable in return string and it does not affect anything

app/main.py
```diff
@@ -1 +1,21 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            errors_count += 1
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+            errors_count += 1
+
+    if masks_to_buy > 0:
```
Y-Havryliv, line 19:
```suggestion
    if masks_to_buy:
```

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from typing import List
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: List[dict], cafe: Cafe) -> str:
+    masks_to_buy = 0
+    errors_count = 0
```
Y-Havryliv, line 8: why do you need this variable?

app/cafe.py
```diff
@@ -0,0 +1,22 @@
+import datetime
```

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from typing import List
```
Y-Havryliv, line 1: add line break here

Y-Havryliv (changes requested):

YuriiKindrat (approved):

PR #333, opened 2023-02-19
Solution

myroslavatrschk:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+
+from app.errors import NotVaccinatedError
+from app.errors import NotWearingMaskError
+from app.errors import OutdatedVaccineError
```
destr34, line 5: format should be like this from app.errors import (NotVaccinatedError,
                                                                                     NotWearingMaskError) 
check readme_file check_list!

app/main.py
```diff
@@ -1 +1,17 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for person in friends:
+        try:
+            Cafe.visit_cafe(cafe, person)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy != 0:
```
destr34, line 15: check example check_list

destr34 (changes requested):

destr34 (approved):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (NotWearingMaskError,
+                        VaccineError)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for person in friends:
```
Y-Havryliv, line 8: I would add some line breaks before and after the for-loop to make your code more readable

app/main.py
```diff
@@ -1 +1,17 @@
...
+        try:
+            Cafe.visit_cafe(cafe, person)
```
Y-Havryliv, line 10: why do you call the method using the class name?

app/main.py
Y-Havryliv: yeah, re-read checklist

Y-Havryliv (changes requested):

YuriiKindrat (approved):

PR #336, opened 2023-02-19
Solution

dogmoony:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> None:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            Cafe.visit_cafe(self=cafe, visitor=friend)
```
ZabFTFT, line 9: Kate, why did you call visit_cafe method with class object and then put the instance with self key. Its up to you, just interesting =)

ZabFTFT (approved):
Well done! But you stunned me with method call =)))

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
```

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Cafe:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("NotVaccinatedError in line 13")
```
Y-Havryliv, line 14: this error message is not good enough, why do you need the line number and what's the point of the error message to be the same as the error name? fix everywhere

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> None:
+    masks_to_buy = 0
+    for friend in friends:
```
Y-Havryliv, line 7: I would add some line breaks before and after the for-loop to make your code more readable

app/main.py
Y-Havryliv: no , it's not up to you in this case. The correct way is `cafe.visit_cafe(friend)`

app/main.py
```diff
@@ -1 +1,16 @@
...
+        try:
+            Cafe.visit_cafe(self=cafe, visitor=friend)
+        except VaccineError:
+            return ("All friends should be vaccinated")
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy:
+        return (f"Friends should buy {masks_to_buy} masks")
```
Y-Havryliv, line 15: why do you write it inside parentheses?

app/main.py
```diff
@@ -1 +1,16 @@
...
+    return (f"Friends can go to {cafe.name}")
```
Y-Havryliv, line 16: same here

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

PR #339, opened 2023-02-19
Solution

StreetNik:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+    @property
+    def name(self) -> str:
+        return self._name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("NotVaccinatedError")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        if not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 23: use .get()

ArsenPidhoretskyi: because wearing_a_mask may not exist

ArsenPidhoretskyi (changes requested):

app/cafe.py
Grayder0152: Pls, do this

app/main.py
```diff
@@ -1 +1,17 @@
...
+from app.cafe import Cafe
+from app.errors import NotWearingMaskError, VaccineError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = len(friends)
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+            masks_to_buy -= 1
```
Grayder0152, line 10: May better set `masks_to_buy` as `0` and when excepted `NotWearingMaskError`  add +1

app/cafe.py
```diff
@@ -0,0 +1,25 @@
+import datetime
+
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
```
Grayder0152, line 7: Pls, order imports

Grayder0152 (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+    @property
+    def name(self) -> str:
+        return self._name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("NotVaccinatedError")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        if not visitor.get("wearing_a_mask"):
```
Grayder0152, line 23: More correct: not visitor.get("wearing_a_mask", False)

Grayder0152 (approved):

Grayder0152 (approved):

PR #345, opened 2023-02-21
Solution

Roman28101:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+import datetime
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
```
Grayder0152, line 5: pls order imports

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Must be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Vaccine expired")
+        if not visitor["wearing_a_mask"]:
```
Grayder0152, line 19:
```suggestion
        if not visitor.get("wearing_a_mask", False):
```
as `visitor` can be without key `wearing_a_mask`

Grayder0152 (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #346, opened 2023-02-22
Solution

lotrik84:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+
+from app.errors import (NotVaccinatedError,
```
vasylhnatiuk, line 3: it would be better to use 
```
from app.errors import (
    NotVaccinatedError,
    NotWearingMaskError,
    OutdatedVaccineError,
)
```

vasylhnatiuk (approved):
Well done!

app/cafe.py
lotrik84: ok, thanks.

PR #349, opened 2023-02-22
Solution

WalDragnir:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if visitor.get("vaccine") is None:
+            raise NotVaccinatedError("Need Vaccine doze")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Do new vaccine doze")
+        elif not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 16: use .get because wearing_a_mask may not exists

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (VaccineError, NotWearingMaskError)
```
katevasilenko, line 2: The brackets is redundant here

katevasilenko (changes requested):
Only one picky remark

vasylhnatiuk (approved):
Well done!

PR #352, opened 2023-02-27
Solution

yhwh6:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError. "
+                                     "Vaccination is not confirmed")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("OutdatedVaccineError. "
+                                       "The vaccine has expired")
+        elif visitor["wearing_a_mask"] is False:
```
ArsenPidhoretskyi, line 20: wearing_a_mask may not exist in visitor
use if visitor.get("wearing_a_mask"):

ArsenPidhoretskyi (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Cafe:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError. "
```
ArsenPidhoretskyi, line 15: Remove text like NotVaccinatedError. we will already know it due to raise type error

ArsenPidhoretskyi (changes requested):
minor comment

PR #357, opened 2023-03-02
Solution

khachatrian2000:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("The visitor is not vaccinated.")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("The vaccine is expired")
+        if not visitor["wearing_a_mask"]:
```
Bohdan-Salamakha, line 19: use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("The visitor is not vaccinated.")
+        if visitor["vaccine"]["expiration_date"] < date.today():
```
Bohdan-Salamakha, line 17: use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

Bohdan-Salamakha (changes requested):
v1

vasylhnatiuk (approved):
Good job!

PR #362, opened 2023-03-10
Solution

kostya-kononenko:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+import datetime
+
+from app.errors import OutdatedVaccineError, \
```
vasylhnatiuk, line 3: please use () for many imports

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+from app.errors import NotWearingMaskError, VaccineError
+
+
+def go_to_cafe(friends: list, cafe: object) -> str:
+    count = 0
```
vasylhnatiuk, line 5: count of what? please add more accurate name

vasylhnatiuk (changes requested):

Grayder0152 (approved):

PR #364, opened 2023-03-13
Solo

IgoraStar:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        masks_to_buy = 1
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("OutdatedVaccine")
+        if visitor["wearing_a_mask"] is False:
```
ArsenPidhoretskyi, line 16: wearing_a_mask may not exist in visitor

ArsenPidhoretskyi (changes requested):

app/cafe.py
```diff
@@ -13,9 +13,10 @@ def visit_cafe(self, visitor: dict) -> str:
             raise NotVaccinatedError("All friends should be vaccinated")
         if visitor["vaccine"]["expiration_date"] < datetime.date.today():
             raise OutdatedVaccineError("OutdatedVaccine")
-        if visitor["wearing_a_mask"] is False:
-            raise NotWearingMaskError(
-                f"Friends should buy {masks_to_buy} masks"
-            )
+        if "wearing_a_mask" in visitor:
```
vasylhnatiuk, line 16: use `get` method https://www.w3schools.com/python/ref_dictionary_get.asp instead of this code

app/cafe.py
```diff
@@ -0,0 +1,22 @@
+import datetime
+from app.errors import NotVaccinatedError,\
```
vasylhnatiuk, line 2: use () instead of \ for many imports

vasylhnatiuk (approved):
Well done!

PR #367, opened 2023-03-27
Solution

MaksNochvai:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(f"{self.name}, you're not vaccinated!")
+        if date.today() > visitor.get("vaccine").get("expiration_date"):
```
olena-hul, line 16:
```suggestion
        if date.today() > visitor.get("vaccine", {}).get("expiration_date"):
```

This will be safer. But unfortunately, Type Error may still occur if expiration_date is None
This is impossible in this task, but generally it is a common scenario

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+            raise OutdatedVaccineError(f"{self.name} "
+                                       f"your vaccination is overdue!")
+        if not visitor["wearing_a_mask"]:
```
olena-hul, line 19:
```suggestion
        if not visitor.get("wearing_a_mask"):
```

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #368, opened 2023-03-28
'Solution'

dominyevhen:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor not vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Vaccine of visitor is outdated")
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("Visitor not wearing mask")
+        else:
+            return f"Welcome to {self.name}"
```
Bohdan-Salamakha, line 18: remove redundant else

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```

app/main.py
```diff
@@ -1 +1,20 @@
...
+        except VaccineError:
+            friends_without_vaccine += 1
+        except NotWearingMaskError:
+            friends_without_mask += 1
+    if friends_without_vaccine > 0:
+        return "All friends should be vaccinated"
+    if friends_without_vaccine == 0 and friends_without_mask > 0:
+        return f"Friends should buy {friends_without_mask} masks"
+    if friends_without_vaccine == 0 and friends_without_mask == 0:
+        return f"Friends can go to {cafe.name}"
```
Bohdan-Salamakha, line 20: u don't need to check the last condition

Bohdan-Salamakha (changes requested):
v1

PR #369, opened 2023-03-28
Solution

AlexandrZhezheria:


app/cafe.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def visit_cafe(
+            self,
+            visitor: dict[str, str | int | bool | dict[str, date]]
+    ) -> str | NotVaccinatedError | OutdatedVaccineError | NotWearingMaskError:
+        if visitor.get("vaccine") is None:
+            raise NotVaccinatedError(
+                f"{visitor.get('name')} is not vaccinated"
+            )
+        if date.today() > visitor.get("vaccine").get("expiration_date"):
```
ArsenPidhoretskyi, line 23: In this case .get is not needed
visitor["vaccine"]["expiration_date"]

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #373, opened 2023-04-01
visitors are verified

dirolius:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("Visitor isn't vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("the vaccine for visitors is outdated")
+        if visitor["wearing_a_mask"] is False:
```
Bohdan-Salamakha, line 16: use "is" only to compare value with None

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+            raise NotWearingMaskError("visitors without masks")
+
+        else:
```
Bohdan-Salamakha, line 19: remove redundant else

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #375, opened 2023-04-04
'Solution'

contritum:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("NotVaccinatedError")
+        elif date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        elif visitor["wearing_a_mask"] is False:
```
ArsenPidhoretskyi, line 16: use .get because wearing_a_mask may not exist in visitor
also None is equal to visitor not have mask

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    count = 0
+    count_masks = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            count += 1
```
ArsenPidhoretskyi, line 12: You should return here. If someone not vaccinate no need to check others

app/main.py
```diff
@@ -1 +1,22 @@
...
+        except NotWearingMaskError:
+            count_masks += 1
+    if count > 0:
+        return "All friends should be vaccinated"
+    elif count_masks > 0:
+        return f"Friends should buy {count_masks} masks"
+    else:
+        if isinstance(cafe, str):
+            return f"Friends can go to {cafe}"
+        return f"Friends can go to {cafe.name}"
```
ArsenPidhoretskyi, line 22: it always accepts cafe no need isinstance

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    count = 0
+    count_masks = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            count += 1
```
ArsenPidhoretskyi, line 15: @contritum not fixed no need to calculate "count"

ArsenPidhoretskyi:
```suggestion
            return "All friends should be vaccinated"
```

ArsenPidhoretskyi (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("NotVaccinatedError")
+        elif date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        elif visitor.get("wearing_a_mask") in [None, False]:
```
EdAlekseiev, line 18: You can use `not visitor.get("wearing_a_mask")` instead of `visitor.get("wearing_a_mask") in [None, False]`

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
```
EdAlekseiev, line 14: You can use `"vaccine" not in visitor`

EdAlekseiev (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError")
+        elif date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        elif not visitor.get("wearing_a_mask"):
+            raise NotWearingMaskError("NotWearingMaskError")
+        else:
+            return f"Welcome to {self.name}"
```
Bohdan-Salamakha, line 21: provide more informatically errors text like "Visitor should be vaccinated" 
use "if" instead of "elif" cause in a case of exceptions raising u will be thrown out of function
remove redundant "else"

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (
+    VaccineError,
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
Bohdan-Salamakha, line 8: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

Bohdan-Salamakha (changes requested):
v1

app/cafe.py
contritum: Ok, I agree. But errors text don't determine in readme. 

Записати

ArsenPidhoretskyi (approved):
Well done!

PR #376, opened 2023-04-05
'Solution'

OleksandrYanchuk:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("The visitor is not vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("The vaccine is expired")
+        if not visitor["wearing_a_mask"]:
```
olena-hul, line 18: Use `.get()` here

olena-hul (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("Visitor isn't vaccinated!")
+
+        if visitor.get("vaccine").get("expiration_date") \
+                < \
+                datetime.date.today():
```
ArsenPidhoretskyi, line 19: So bad readability.... Also .get for vaccine is not needed. And for expiration date also, because if vaccine present we always have expiration_date
```suggestion
        expiration_date = visitor["vaccine"]["expiration_date"]
        if expiration_date < datetime.date.today():
```

ArsenPidhoretskyi (changes requested):

PR #377, opened 2023-04-05
Solution

goldenuni:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
+import datetime
+import app.errors
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if visitor.get("vaccine", 0):
```
ArsenPidhoretskyi, line 10: Why you need 0 as default? Be default it return None

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if visitor.get("vaccine", 0):
+            if visitor["vaccine"]["expiration_date"] >= datetime.date.today():
+                if not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 12: wearing_a_mask may not exist inside visitor. Use .get here

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #384, opened 2023-04-14
Solution Lockdown Cafe

vborovets0:


app/cafe.py
```diff
@@ -0,0 +1,17 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError
+        if not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 15: use .get because wearing_a_mask may not exist

ArsenPidhoretskyi (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,17 @@
+import datetime
+from app.errors import NotVaccinatedError, OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
Bohdan-Salamakha, line 3: unite these imports

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```

Bohdan-Salamakha (changes requested):
v1

app/cafe.py
```diff
@@ -0,0 +1,17 @@
+import datetime
+from app.errors import (NotVaccinatedError, OutdatedVaccineError,
+                        NotWearingMaskError)
```
Bohdan-Salamakha, line 3:
```
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```

app/main.py
```diff
@@ -1 +1,21 @@
...
+        except NotWearingMaskError:
+            count += 1
+
+        except VaccineError:
+            return "All friends should be vaccinated"
+
+    if count > 0:
+        return f"Friends should buy {count} masks"
+
+    return f"Friends can go to {cafe.name}"
```
Bohdan-Salamakha, line 21: u don't need so many newlines in your code
newlines should separate different logic blocks
u don't have so different logic to use newlines as separators

Bohdan-Salamakha (changes requested):
v3

Bohdan-Salamakha (approved):

PR #387, opened 2023-04-21
Solution

antonpiddubnyi:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
+import datetime
+from app.errors import \
+    NotVaccinatedError, \
+    OutdatedVaccineError, \
+    NotWearingMaskError
+
```
Grayder0152, line 6: use `()` instead of `\` and sort imports

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from app.errors import \
+    NotVaccinatedError, \
+    OutdatedVaccineError, \
+    NotWearingMaskError
```
Grayder0152, line 4: the same as above

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (
+    NotVaccinatedError,
+    NotWearingMaskError,
+    OutdatedVaccineError
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
ArsenPidhoretskyi, line 9: friends: list of what?

app/main.py
```diff
@@ -1 +1,20 @@
...
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
ArsenPidhoretskyi, line 14: You can catch parent exception instead of two

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("Please vaccinate yourself")
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Your vaccine is expired")
+
+        if not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 20: wearing_a_mask may not exist inside visitor. Use .get method

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #389, opened 2023-04-24
Solution

MaksymUK:


app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+            raise NotVaccinatedError("NotVaccinatedError")
+
+        today_date = datetime.date.today()
+        if today_date > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("OutdatedVaccineError")
+
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("NotWearingMaskError")
+
+        if no_exception:
```
olena-hul, line 25: This check makes no sense, just return the result without `if`

app/cafe.py
```diff
@@ -0,0 +1,26 @@
+import datetime
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
```
olena-hul, line 4:
```suggestion
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```

app/main.py
```diff
@@ -1 +1,22 @@
...
+            return "All friends should be vaccinated"
+
+        except NotWearingMaskError:
+            masks_to_buy += 1
+            all_allowed = False
+
+    if all_allowed:
+        return f"Friends can go to {cafe.name}"
+    else:
+        return f"Friends should buy {masks_to_buy} masks"
```
olena-hul, line 22: You don't need to use additional `all_allowed` variable here. You can just check if `masks_to_buy > 0`

olena-hul: Moreover, no need to use `else` keyword after `return`

olena-hul (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError")
+
+        today_date = datetime.date.today()
+        if today_date > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("OutdatedVaccineError")
+
+        if not visitor.get("wearing_a_mask"):
+            raise NotWearingMaskError("NotWearingMaskError")
```
Bohdan-Salamakha, line 24: provide more detailed exception description like "Customer should be vaccinated"

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #390, opened 2023-04-28
Solution

KittNeKit:


Bohdan-Salamakha (changes requested):
don't work on the master branch - it's a bad practice
https://github.com/mate-academy/py-task-guideline/blob/main/README.md#:~:text=git%20checkout%20%2Db%20develop

PR #391, opened 2023-04-28
Solution

KittNeKit:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("All friends should be vaccinated")
+        if visitor["wearing_a_mask"] is False:
```
Bohdan-Salamakha, line 16: u don't need to use id comparison in this case
```suggestion
        if not visitor["wearing_a_mask"]:
```

app/cafe.py
```diff
@@ -0,0 +1,18 @@
+from datetime import date
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
+from app.errors import NotVaccinatedError
```
Bohdan-Salamakha, line 4: unite these imports

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.errors import VaccineError
+from app.errors import NotWearingMaskError
```
Bohdan-Salamakha, line 2: unite these imports

app/main.py
```diff
@@ -1 +1,17 @@
...
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```

Bohdan-Salamakha (changes requested):
v1

app/cafe.py
```diff
@@ -0,0 +1,17 @@
+from datetime import date
+from app.errors import OutdatedVaccineError, NotVaccinatedError
+from app.errors import NotWearingMaskError
```
Bohdan-Salamakha, line 3:
```suggestion
from app.errors import (
    OutdatedVaccineError,
    NotVaccinatedError,
    NotWearingMaskError
)
```

Bohdan-Salamakha (changes requested):
v3

Bohdan-Salamakha (approved):

PR #393, opened 2023-05-01
Solution

VladyslavKaminskyi:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        if visitor["wearing_a_mask"] is False:
```
ArsenPidhoretskyi, line 18: wearing_a_mask may not exist in visitor

ArsenPidhoretskyi (changes requested):

PR #395, opened 2023-05-02
Solution

Dimdimior:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError
+        if visitor["vaccine"].get("expiration_date") < datetime.date.today():
+            raise OutdatedVaccineError
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError
+        else:
```
Bohdan-Salamakha, line 20: remove redundant "else"

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```

Bohdan-Salamakha (changes requested):
v1

PR #399, opened 2023-05-06
Solution

neostyle88:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```

app/main.py
```diff
@@ -1 +1,17 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks += 1
+    if not masks:
+        return f"Friends can go to {cafe.name}"
+    else:
```
Bohdan-Salamakha, line 16: remove redundant return

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #403, opened 2023-05-10
Solution

markus2812:


app/main.py
```diff
@@ -1 +1,19 @@
...
+    OutdatedVaccineError, NotWearingMaskError
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotVaccinatedError as error_info:
```
vasylhnatiuk, line 11:
```suggestion
        except VaccineError:
```

vasylhnatiuk, line 11:
```suggestion
        except VaccineError:
```

app/main.py
```diff
@@ -1 +1,19 @@
...
+            return str(error_info)
```
vasylhnatiuk, line 12:
```suggestion
           return "All friends should be vaccinated"
```

app/main.py
vasylhnatiuk: It would be better to use VaccineError instead of the two exceptions that inherit from VaccineError.

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
fix it

ArsenPidhoretskyi (approved):
Well done!

PR #415, opened 2023-05-18
solution

IvanKorshunovE:


app/cafe.py
```diff
@@ -0,0 +1,26 @@
+import datetime
+
+from app.errors import NotVaccinatedError, OutdatedVaccineError, \
+    NotWearingMaskError
```
Grayder0152, line 4: Use`()` instead of `\`

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+            OutdatedVaccineError | NotWearingMaskError:
+
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+
+        current_date = datetime.datetime.now().date()
+        if visitor["vaccine"]["expiration_date"] < current_date:
+            raise OutdatedVaccineError
+
+        if not visitor["wearing_a_mask"]:
```
Grayder0152, line 23: use `get` method instead of `[]` as `wearing_a_mask` can does not exist

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,30 @@
...
+                     | NotVaccinatedError
+                     | OutdatedVaccineError
+                     | NotWearingMaskError):
+
+    not_wearing_mask_error = None
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError) as e:
```
vasylhnatiuk, line 19:
```suggestion
        except VaccineError as e:
```

Grayder0152 (approved):

PR #416, opened 2023-05-19
'Solution'

Artistgatehub:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: str) -> str:
```

app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+from app.errors import (
+    NotVaccinatedError, NotWearingMaskError, OutdatedVaccineError
+)
```
Bohdan-Salamakha, line 4:
```suggestion
from app.errors import (
    NotVaccinatedError,
    NotWearingMaskError,
    OutdatedVaccineError
)
```

Bohdan-Salamakha (changes requested):

YukihiroSM (approved):

PR #423, opened 2023-05-29
implemented cafe visit

S-SMTN:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+from app.errors import NotVaccinatedError,\
+    OutdatedVaccineError,\
+    NotWearingMaskError
```
olena-hul, line 4: Use `()` to split lines

Bohdan-Salamakha (approved):

PR #424, opened 2023-05-30
Solution

gleblitvinenko:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("Visitor should be vaccinated")
+        elif datetime.date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("Vaccine is outdated")
+        elif not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 18: wearing_a_mask may not exit inside visitor. Use .get method

app/errors.py
```diff
@@ -0,0 +1,16 @@
+class VaccineError(Exception):
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
```
ArsenPidhoretskyi, line 3: This is not needed. Because it works by default

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (changes requested):
image
fix tests

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("Visitor should be vaccinated")
+        elif (
+                datetime.date.today()
+                > visitor.get("vaccine").get("expiration_date")
```
ArsenPidhoretskyi, line 18: But here .get is not needed. We know that vaccine exist. And if vaccine exits expiration_date also exists

ArsenPidhoretskyi (approved):
Good job!

PR #425, opened 2023-06-01
Solution

yuro740:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("visitor haven't vaccine")
+        elif visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("visitor have outdated vaccine")
+        elif visitor.get("wearing_a_mask") is False:
```
ArsenPidhoretskyi, line 17:
```suggestion
        elif visitor.get("wearing_a_mask", False) is False:
```
better to do like this because None is False -> True

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #437, opened 2023-06-08
Develop

ant-komarov:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
+import datetime
```
AntonShpakovych, line 1: You need to add space between different import group

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine", False):
```
AntonShpakovych, line 12: You don't need to use get (key, False) because get use None for defolt. And even more u can rewrite this to
if not "vaccine" in visitor

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+            raise NotVaccinatedError("All visitors must be vaccinated!!!")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Vaccine outdated!!!")
+        if not visitor.get("wearing_a_mask", False):
```
AntonShpakovych, line 16: First of all u don't need to use if here,  just use elif also key wearing_a_mask all time present but he has different values True or False. So u need rewrite this:
 elif not visitor["wearing_a_mask"]

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine", False):
+            raise NotVaccinatedError("All visitors must be vaccinated!!!")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
AntonShpakovych, line 14: same here don't need to use if

app/main.py
```diff
@@ -1 +1,19 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except NotWearingMaskError:
+            without_mask += 1
+        except VaccineError:
+            return "All friends should be vaccinated"
+
+    if without_mask:
+        return f"Friends should buy {without_mask} masks"
+
```
AntonShpakovych, line 18: don't need space here

app/cafe.py
```diff
@@ -0,0 +1,18 @@
+import datetime
+from app.errors import (NotVaccinatedError,
```
AntonShpakovych, line 2: bad formatting

app/main.py
```diff
@@ -1 +1,19 @@
...
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    without_mask = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotWearingMaskError:
```
AntonShpakovych, line 11: change order for except because VaccineError have bigger priority in this task

AntonShpakovych (changes requested):

app/main.py
ant-komarov: OK, I`ll do that, but if VaccineError is occure, it will be raised first in visit_cafe. It is only handles here.

AntonShpakovych (approved):
Good job!

AntonShpakovych (approved):

PR #443, opened 2023-06-09
'Solution'

RedCatHot:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
+from datetime import date
+from app.errors import (NotWearingMaskError,
+                        NotVaccinatedError,
+                        OutdatedVaccineError)
+
```
IhorVoskoboinikov, line 5: Edit imports: add an empty line between them and make error imports like:
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)

There are no errors in the code block, it just seems to me that it will be more beautiful)))

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.errors import (VaccineError, NotWearingMaskError)
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
IhorVoskoboinikov, line 5: Add annotations likes as list[dict] or List[dict] import from typing -> List

IhorVoskoboinikov (changes requested):
Good job! Make small changes)

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.errors import (VaccineError, NotWearingMaskError)
```
Altahoma, line 1: probably no needed use brackets here if don't wtire each import in individual line

Altahoma (approved):

app/main.py
YuriiKindrat: fix imports according to pep8
https://peps.python.org/pep-0008/#imports:~:text=Imports%20should%20be,group%20of%20imports.

YuriiKindrat (changes requested):

IhorVoskoboinikov (approved):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+from app.cafe import Cafe
+from typing import List
```
YuriiKindrat, line 3: u didn't fix that
https://peps.python.org/pep-0008/#imports:~:text=Imports%20should%20be,group%20of%20imports.

app/cafe.py
```diff
@@ -0,0 +1,20 @@
+from datetime import date
+from app.errors import (
+    NotWearingMaskError,
+    NotVaccinatedError,
+    OutdatedVaccineError
+)
```
YuriiKindrat, line 6: same is here, u should put a blank line between different groups of imports

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
...
+from app.errors import (
+    VaccineError,
+    NotWearingMaskError
+)
+from app.cafe import Cafe
+
+from typing import List
+
+
+def go_to_cafe(friends: List[dict], cafe: Cafe) -> str:
```
LiudmylaKulinchenko, line 10: You can use list[] instead of List in your type hint, even if you're using Python 3.9 or later

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from app.errors import (
+    VaccineError,
+    NotWearingMaskError
+)
+from app.cafe import Cafe
+
+from typing import List
```
LiudmylaKulinchenko, line 7: Mind imports order
Local imorts should be after third-party imports

LiudmylaKulinchenko (changes requested):

PR #445, opened 2023-06-09
'Solution'

yakhlamova:


app/main.py
```diff
@@ -1 +1,24 @@
...
+                        )
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
kovaliskoveronika, line 14: Here can be used VaccineError instead this two

app/main.py
```diff
@@ -1 +1,24 @@
...
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy = sum(
+                1 for friend in friends if not friend["wearing_a_mask"]
+            )
```
kovaliskoveronika, line 19: There is no need to use comprehension, you have already checked if friend has a mask in visit_cafe. Maybe it simpler will be to add one if we raise a NotWearingMaskError error

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+                        )
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
```
kovaliskoveronika, line 14: It is not a mistake, but you can use get() to know if a key is in dict
visitor.get("vaccine")

kovaliskoveronika (changes requested):

app/cafe.py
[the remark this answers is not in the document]
yakhlamova: Yes, you`re right but I like it the way I wrote it. If it`s ok, I prefer to stay with my variant but if it`s a bad practice I will be more than happy to change it.

kovaliskoveronika (approved):

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError
+                        )
```
YuriiKindrat, line 5: bad formatting here(

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (NotWearingMaskError,
+                        VaccineError
+                        )
```
YuriiKindrat, line 4: same

YuriiKindrat (changes requested):

PR #446, opened 2023-06-09
Implemented vaccine errors, Cafe and go_to_cafe()

Altahoma:


app/errors.py
```diff
@@ -0,0 +1,20 @@
...
+
+class NotVaccinatedError(VaccineError):
+    def __init__(self) -> None:
+        self.message = "The visitor does not have a vaccine."
+        super().__init__(self.message)
+
+
+class OutdatedVaccineError(VaccineError):
+    def __init__(self) -> None:
+        self.message = "The vaccine must not be expired."
```
OlhaTryhub, line 13: You don't really need this row. You can just put the text of message directly to super init.

app/errors.py
```diff
@@ -0,0 +1,20 @@
+class VaccineError(Exception):
+    pass
+
+
+class NotVaccinatedError(VaccineError):
+    def __init__(self) -> None:
+        self.message = "The visitor does not have a vaccine."
```
OlhaTryhub, line 7: The same to next comment ☘

app/errors.py
```diff
@@ -0,0 +1,20 @@
...
+class OutdatedVaccineError(VaccineError):
+    def __init__(self) -> None:
+        self.message = "The vaccine must not be expired."
+        super().__init__(self.message)
+
+
+class NotWearingMaskError(Exception):
+    def __init__(self) -> None:
+        self.message = "The visitor must wear a face mask."
+        super().__init__(self.message)
```
OlhaTryhub, line 20: The same to the previous one 🍬

OlhaTryhub (changes requested):

app/errors.py
Altahoma: It's a common practice to store the exception message as an instance variable.
 You can see this by following the link attached to the first lesson in the paragraph ["Customizing Exception Classes"](https://www.programiz.com/python-programming/user-defined-exception#class-customize)

OlhaTryhub: OK, let it be. But I don't really see big sense in this line.

OlhaTryhub (approved):

YuriiKindrat (approved):

PR #447, opened 2023-06-09
Lockdown-Cafe (first commit)

IhorVoskoboinikov:


YKonoplyov (changes requested):

app/main.py
```diff
@@ -1 +1,24 @@
...
+            no_vaccine = True
+            break
+        except NotWearingMaskError:
+            no_mask += 1
+
+    if no_vaccine:
+        return "All friends should be vaccinated"
+    elif no_mask:
+        return f"Friends should buy {no_mask} masks"
+    return "Friends can go to KFC"
```
Kasatello, line 24: there should be `{cafe.name}` variable

Kasatello (changes requested):

Kasatello (approved):

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from __future__ import annotations
+
+
```
YuriiKindrat, line 3: do u really need this here? 
also why do u need 2 blank lines here?

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> Optional[str]:
```
YuriiKindrat, line 15: u can just write str here instead of optional str ))

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🌼

PR #448, opened 2023-06-09
solution

oleksiikolii:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+import datetime
+from app.errors import (NotVaccinatedError,
+                        NotWearingMaskError,
+                        OutdatedVaccineError)
```
YuriiKindrat, line 4: according to pep8 it is better to place a blank line between different groups of imports)

YuriiKindrat (approved):

PR #449, opened 2023-06-09
My solution

kovaliskoveronika:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
+import datetime
+
+from app.errors import NotVaccinatedError
+from app.errors import NotWearingMaskError
+from app.errors import OutdatedVaccineError
```
YuriiKindrat, line 5: could u import errors in one import ?)

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError(
+                "NotVaccinatedError should be raised with a message")
```
YuriiKindrat, line 15: bad formatting here

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+        current_date = datetime.date.today()
+        if visitor["vaccine"]["expiration_date"] < current_date:
+            raise OutdatedVaccineError(
+                "OutdatedVaccineError should be raised with a message")
+        if not visitor.get("wearing_a_mask"):
+            raise NotWearingMaskError(
+                "NotWearingMaskError should be raised with a message")
```
YuriiKindrat, line 22: bad formatting here

app/cafe.py
YuriiKindrat, line 15: it seems to me that this is not descriptive message for user, the same is for other exceptions

app/errors.py
```diff
@@ -0,0 +1,22 @@
+class VaccineError(Exception):
+    def __init__(self, mess: str = None) -> None:
+        self.mess = mess
+
+    def __str__(self) -> str:
+        return self.mess
```
YuriiKindrat, line 6: why do u need this ? 
also mess is a bad name for attr

app/errors.py
```diff
@@ -0,0 +1,22 @@
...
+class OutdatedVaccineError(VaccineError):
+    pass
+
+
+class NotWearingMaskError(Exception):
+    def __init__(self, mess: str = None) -> None:
+        self.mess = mess
+
+    def __str__(self) -> str:
+        return self.mess
```
YuriiKindrat, line 22: same is here

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,24 @@
...
+                        )
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
YuriiKindrat, line 14: also could u use parent class here ?

YuriiKindrat (changes requested):

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #450, opened 2023-06-09
Solution

Gvilion:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("You should be vaccinated")
+        if ((visitor["vaccine"]["expiration_date"]
+             - datetime.today().date()).days < 0):
```
Po-SvinSkipodla, line 18: you can compare only two dates

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
```
Po-SvinSkipodla, line 7: add VaccineError to error import

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    without_mask_count = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
Po-SvinSkipodla, line 15: and use it here, instead of 2 errors

Po-SvinSkipodla (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            without_mask_count += 1
+
+    if without_mask_count > 0:
```
LiudmylaKulinchenko, line 19: it is okay, but u can just use `if masks_to_buy:` , since it will return True only if masks_to_buy > 0

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("You should be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.today().date():
+            raise OutdatedVaccineError("Yore vaccine should be valid")
```
LiudmylaKulinchenko, line 18: Can fix the typo at the same time 😋

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+        if not visitor.get("wearing_a_mask"):
+            raise NotWearingMaskError("You should have a mask")
```
LiudmylaKulinchenko, line 20: Also it will be more correctly "You should wear a mask"

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🚀

PR #455, opened 2023-06-10
Solution

Po-SvinSkipodla:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("NotWearingMaskError")
```
YuriiKindrat, line 19: add more descriptive error messages here
no need to repeat an exception name in message

app/cafe.py
```diff
@@ -0,0 +1,21 @@
+import datetime
+from app.errors import (
+    NotWearingMaskError,
+    NotVaccinatedError,
+    OutdatedVaccineError
+)
```
YuriiKindrat, line 6: add blank line between 2 different groups of imports

app/main.py
```diff
@@ -1 +1,23 @@
...
+                cafe.visit_cafe(friend)
+            except NotWearingMaskError:
+                masks_to_buy += 1
+        if masks_to_buy:
+            raise NotWearingMaskError
+    except NotWearingMaskError:
+        return f"Friends should buy {masks_to_buy} masks"
+    except VaccineError:
+        return "All friends should be vaccinated"
+    else:
```
YuriiKindrat, line 22: why do u need else here? )

app/main.py
```diff
@@ -1 +1,23 @@
...
+            try:
+                cafe.visit_cafe(friend)
+            except NotWearingMaskError:
+                masks_to_buy += 1
+        if masks_to_buy:
+            raise NotWearingMaskError
+    except NotWearingMaskError:
+        return f"Friends should buy {masks_to_buy} masks"
+    except VaccineError:
+        return "All friends should be vaccinated"
```
YuriiKindrat, line 21: could u write this using only one try block ?

YuriiKindrat (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,23 @@
+import datetime
+
+
+from app.errors import (
+    NotWearingMaskError,
+    NotVaccinatedError,
+    OutdatedVaccineError
+)
```
Oleg-s-s, line 8: I think 1 line would be enough tho

Oleg-s-s (approved):

PR #456, opened 2023-06-11
Solution

DmytroTsehelnyk:


Mr-Furman (approved):
As for me, GJ

app/cafe.py
```diff
@@ -0,0 +1,28 @@
...
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        vaccine = visitor.get("vaccine")
+        wearing_a_mask = visitor.get("wearing_a_mask")
```
YuriiKindrat, line 16: no need to create var if u are using it only once

app/main.py
```diff
@@ -1 +1,21 @@
...
+        try:
+            cafe.visit_cafe(person)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_counter += 1
+
+    if not masks_counter:
+        return f"Friends can go to {cafe.name}"
+    return f"Friends should buy {masks_counter} masks"
```
YuriiKindrat, line 21: could u use backward condition here, to write less code ?

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from app.cafe import Cafe
+
+from app.errors import (
+    VaccineError,
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(friends: list, cafe: "Cafe") -> str:
```
YuriiKindrat, line 9: why do u need quotes for Cafe ?

YuriiKindrat (changes requested):

app/main.py
DmytroTsehelnyk: I can make backward condition, but it still will be 3 lines of code

app/main.py
```diff
@@ -1 +1,21 @@
...
+    masks_counter = 0
+    for person in friends:
+        try:
+            cafe.visit_cafe(person)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_counter += 1
+
+    if masks_counter > 0:
```
YuriiKindrat, line 19: it is okay, but u can just write `if masks_counter` here)

YuriiKindrat (changes requested):

PR #457, opened 2023-06-11
Solution lockdown cafe

MarkShapka:


app/main.py
```diff
@@ -1 +1,19 @@
...
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            Cafe.visit_cafe(cafe, friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
```
Dimosphen1, line 15: Consider ending the try-except block with catching and processing general `Exception`. This will help to avoid non-foreseen errors that may arise in the `visit_cafe` function.

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        expiration_date = visitor["vaccine"].get("expiration_date")
+        if expiration_date < datetime.date(datetime.today()):
+            raise OutdatedVaccineError
+        if not visitor["wearing_a_mask"]:
```
Dimosphen1, line 17: In case the developer is sure of the structure of the incoming `visitor` dict, it is safe to leave the value retrieval for the `wearing_a_mask` key as is. In the production code in most cases it is better to check for the presence of a key before retrieving a value, that is why you could consider the safer method of retrieving a value and checking for the presence of the key at the same time.

Dimosphen1 (approved):
Well done, project structure, code style, and solution overall cover requirements.

PR #458, opened 2023-06-12
Solution

YKonoplyov:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+
+
+from app.errors import (NotVaccinatedError,
```
YuriiKindrat, line 4: why do u need 2 blank lines here?

YuriiKindrat (changes requested):

Oleg-s-s (approved):
:+1:

PR #466, opened 2023-06-15
'Solution'

sind14:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```

Bohdan-Salamakha (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("Vaccine is expired")
+
+        if not visitor["wearing_a_mask"]:
```
olena-hul, line 20: Use .get method to avoid KeyError

olena-hul (changes requested):

PR #472, opened 2023-06-22
Solution

NazarLytvyniuk:


app/main.py
```diff
@@ -1 +1,22 @@
...
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    should_be_vaccinated = False
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            should_be_vaccinated = True
```
Dimosphen1, line 12: There is no need for `should_be_vaccinated` boolean flag, in case `VaccineError` is caught there should be a return statement with a required message.

app/cafe.py
```diff
@@ -0,0 +1,23 @@
+import datetime
+
+from app.errors import NotVaccinatedError,\
```
Dimosphen1, line 3: Import each error from `app.errors` on the new line with the use of parenthesis.

Dimosphen1 (changes requested):
Several changes are requested.

app/cafe.py
```diff
@@ -0,0 +1,25 @@
+import datetime
+
+from app.errors import (NotVaccinatedError,
```
Polyakiv-Andrey, line 3: If you have 1 element after the bracket and then transfer to the next line, then in this case this element should also be on a new line

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #483, opened 2023-07-02
'Solution'

Kavbian:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
+import app.errors as er
```
ArsenPidhoretskyi, line 1: Use:
from app.errors import (
    and here all exceptions
)

ArsenPidhoretskyi (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None | str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError
+
+        if not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 16: wearing_a_mask may not exist inside visitor

ArsenPidhoretskyi: Use .get methods

ArsenPidhoretskyi (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from app.errors import NotVaccinatedError, \
+    OutdatedVaccineError, NotWearingMaskError
```
ArsenPidhoretskyi, line 2:
```suggestion
from app.errors import (
    NotVaccinatedError, OutdatedVaccineError, NotWearingMaskError
)
```
Better to use brackets instead of \
Also you can do:
```
from app.errors import (
    NotVaccinatedError, 
    OutdatedVaccineError, 
    NotWearingMaskError
)
```

ArsenPidhoretskyi (approved):
Good job!

PR #491, opened 2023-07-05
Solution

alecilchenko:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError)
```
Bohdan-Salamakha, line 5:
```suggestion
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
    )
```

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from __future__ import annotations
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #492, opened 2023-07-07
Solution

pnakongit:


app/cafe.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(
+            self,
+            visitor: dict
+    ) -> str | VaccineError | NotWearingMaskError:
```
vasylhnatiuk, line 18: this function can raise errors and can not return errors 
```suggestion
    ) -> str:

```

vasylhnatiuk (changes requested):

app/cafe.py
pnakongit: understood. Thx a lot !

app/cafe.py
```diff
@@ -0,0 +1,34 @@
+import datetime
+
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
```
Grayder0152, line 6: sort import

app/cafe.py
```diff
@@ -0,0 +1,34 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(
+            self,
+            visitor: dict
+    ) -> str:
+
+        if not visitor.get("vaccine", None):
```
Grayder0152, line 19:
```suggestion
        if visitor.get("vaccine") is None:

```

Grayder0152 (changes requested):

app/cafe.py
pnakongit: Done. Thx a lot !

app/cafe.py
pnakongit: Done! Thx a lot !

app/cafe.py
```diff
@@ -0,0 +1,34 @@
...
+            raise NotVaccinatedError(
+                "The visitor must be vaccinated"
+            )
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError(
+                "The vaccine has expired"
+            )
+
+        if not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 29: wearing_a_mask may not exist inside visitor. Use .get method

ArsenPidhoretskyi (changes requested):

app/cafe.py
pnakongit: Done. Thanks !

app/cafe.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(
+            self,
+            visitor: dict
+    ) -> str:
```
vsmutok, line 17:
```suggestion
    def visit_cafe(self, visitor: dict) -> str:

```

vsmutok (approved):
Great 👍

PR #493, opened 2023-07-10
- solution

Oleksiy-Lyashenko:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+from app.errors import NotVaccinatedError, \
```
Polyakiv-Andrey, line 1: Use parentheses for hyphenation, not /.

PR #498, opened 2023-07-13
solution

vkleshko:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        datetime_today = datetime.date.today()
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        if visitor["vaccine"]["expiration_date"] < datetime_today:
+            raise OutdatedVaccineError
+        if not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 19: Use .get because wearing_a_mask may not exist inside the visitor:
image

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+            raise NotWearingMaskError
+        else:
+            return f"Welcome to {self.name}"
```
ArsenPidhoretskyi, line 22:
```suggestion
        return f"Welcome to {self.name}"
```
else: is redundant

ArsenPidhoretskyi (changes requested):

PR #499, opened 2023-07-15
Implemented all logic

lubart2517:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError, NotWearingMaskError)
+
```
pavlejviki, line 4:
```suggestion
import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("Visitors must be vaccinated")
+        else:
+            vaccine = visitor["vaccine"]
```
pavlejviki, line 14: No need to create a separate variable if you are using it just once.

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    vaccine_error_counter = 0
+    mask_error_counter = 0
+    for person in friends:
+        try:
+            cafe.visit_cafe(person)
+        except VaccineError:
+            vaccine_error_counter += 1
```
pavlejviki, line 13: You don't need the vaccine_error_counter. You can return here the error message straight away.

app/main.py
```diff
@@ -1 +1,22 @@
...
+        except NotWearingMaskError:
+            mask_error_counter += 1
+
+    if vaccine_error_counter:
```
pavlejviki, line 17: Return the message when the error is caught.

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

Polyakiv-Andrey (approved):
Good job!

PR #503, opened 2023-07-19
Develop

3lobites:


3lobites:
I object to the text of the task, everyone knows that the vaccine is valid for a year. And not day to day, which follows only from the analysis of the test.

YukihiroSM (changes requested):
Fix the comment below, please. Else is good!

PR #505, opened 2023-07-21
Solution

bogdandovgan:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("You should be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("You should be vaccinated")
+        if visitor["wearing_a_mask"] is False:
```
ArsenPidhoretskyi, line 19: wearing_a_mask may not exist. Use .get

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+            raise NotWearingMaskError("You should put on a mask")
+        else:
+            return f"Welcome to {self.name}"
```
ArsenPidhoretskyi, line 22:
```suggestion
        return f"Welcome to {self.name}"
```
else: is not needed

ArsenPidhoretskyi (changes requested):

PR #508, opened 2023-07-26
Solution

nschcur:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+
+import datetime
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
Grayder0152, line 5: sort and group imports

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError
+from app.errors import NotWearingMaskError
```
Grayder0152, line 3: the same as above

Grayder0152 (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #515, opened 2023-08-03
Solution

fotomassimo:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (VaccineError, NotWearingMaskError)
```
fsocie7y, line 2: remove redundant parentheses

app/main.py
```diff
@@ -1 +1,16 @@
...
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    unmasked_count = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
```
fsocie7y, line 10: better to use NotVaccinatedError

app/main.py
fotomassimo: done

app/main.py
fotomassimo: According to the paragraph "Notes" in the "README.md" file, the VaccineError exception should be used to catch both NotVaccinatedError and OutdatedVaccineError that could be thrown by the method "visit_cafe", as per project task. Also, it seems the tests of this task imply this kind of logic otherwise they won't be passed, unless an extra "except" clause will be added to catch OutdatedVaccineError with the same return message which seems to be redundant and doesn't comply with the stated requirements.

PR #517, opened 2023-08-06
'Solution'

niksgawrilenko:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+
+from app.errors import (NotVaccinatedError,
```
Polyakiv-Andrey, line 3: After the opening bracket, write all arguments on a new line, and write the closing bracket on a new line as well.

Polyakiv-Andrey (changes requested):

app/cafe.py
niksgawrilenko: Done

PR #521, opened 2023-08-08
Lockdown cafe

IProskurnytskyi:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | None:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("You have to be vaccinated.")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Your vaccine is expired.")
+        elif not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("You have to be in mask.")
+        return f"Welcome to {self.name}"
```
pavlejviki, line 20: Please use the get() method to verify the presence of a key and to fetch its value, as this approach can prevent KeyError exceptions .

pavlejviki (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | None:
+        if "vaccine" not in visitor:
```
pavlejviki, line 14:
```suggestion
        if not visitor.get("vaccine"):
```

pavlejviki (approved):
Good job!

PR #522, opened 2023-08-09
lockdown-cafe

MishkaGAMII:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today:
+            raise OutdatedVaccineError
+        elif not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 18: wearing_a_mask may not exists inside visitor

ArsenPidhoretskyi: Use .get method

ArsenPidhoretskyi (changes requested):

PR #523, opened 2023-08-10
Solution

AlexPelevin99:


app/main.py
```diff
@@ -1 +1,28 @@
...
+        except NotVaccinatedError as e:
+            return str(e)
+        except OutdatedVaccineError as e:
+            return str(e)
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy:
+        return f"Friends should buy {masks_to_buy} masks"
+    if allow_friend_count == len(friends):
```
Polyakiv-Andrey, line 25: Remove this if and just return Friends can go to {cafe.name}, so as not to write your last return

Polyakiv-Andrey (changes requested):

PR #527, opened 2023-08-15
Solution

eduardhabryd:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError
```
Kroll410, line 17: add descriptive error message

Kroll410 (changes requested):

app/cafe.py
eduardhabryd: > add descriptive error message

I have those descriptions specified in errors.py, please check.

Kroll410 (approved):

Kroll410:
In this case it's okay to specify error message in the exact error, consider about having error messages in place for generic ones

PR #538, opened 2023-08-19
'Solution'

hanovskyiv:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+from app.errors import (NotVaccinatedError, OutdatedVaccineError,
```
Polyakiv-Andrey, line 2: After opening the parentheses, write all the arguments on a new line, including the first one, the closing bracket is also written on a new line

Polyakiv-Andrey (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+        try:
+            if datetime.date.today() > visitor["vaccine"]["expiration_date"]:
+                raise OutdatedVaccineError
+            if not visitor["wearing_a_mask"]:
+                raise NotWearingMaskError
+            return f"Welcome to {self.name}"
+        except KeyError:
+            raise NotVaccinatedError
```
Bohdan-Salamakha, line 22: instead of using try except u can use this check
```
        if "vaccine" not in visitor:
            raise NotVaccinatedError
```

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```

Bohdan-Salamakha (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Cafe:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        if datetime.date.today() > visitor["vaccine"]["expiration_date"]:
```
Dimosphen1, line 17: It would be preferable to place at the first position a value that you want to check from the function input and the value you check against in the second position.

Dimosphen1, line 17: In case you are unsure about the incoming dict structure, it would be preferable to use `.get()` in production settings to retrieve the required values.

Dimosphen1 (approved):
LGTM, well done! Several comments are left for consideration.

PR #551, opened 2023-08-21
solution

sashabryl:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    vaccination_issues = False
```
Dimosphen1, line 7: There is no need for the `vaccination_issues` variable, you can return the required string at the moment of exception processing.

app/main.py
```diff
@@ -1 +1,19 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            vaccination_issues = True
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if vaccination_issues:
+        return "All friends should be vaccinated"
```
Dimosphen1, line 16: Remove this if condition.

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(
+            self,
+            name: str
+    ) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | Exception:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitors must be vaccinated!")
+        if visitor["vaccine"]["expiration_date"] < date.today():
```
Dimosphen1, line 19: Using `.get()` is a safer way of getting a value from a dict. You should use `[]` only when you are sure of the structure of the incoming dict. Generally, in production code prefer using `.get()` over `[]`.

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+            raise OutdatedVaccineError("Your vaccination term is expired!")
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("All visitors must wear a mask!")
+        return "Welcome to " + self.name
```
Dimosphen1, line 23: Prefer using f-string instead of string concatenation, it is more descriptive. Rewrite this line with the use of f-string.

Dimosphen1 (changes requested):

sashabryl:
I made it slightly different with get() method but it's still pretty the same, isn't it? I can't just call visitor.get("expiration_date") because it returns None, as it is nested in another dictionary, I guess, or can I?

Dimosphen1 (approved):
LGTM, well done!

PR #556, opened 2023-08-21
Solution with Exceptions

ZaichenkoViktoriia:


app/cafe.py
```diff
@@ -0,0 +1,23 @@
+import datetime
+from app.errors import (
```

Y-Havryliv (approved):

Y-Havryliv (approved):

PR #558, opened 2023-08-21
Solution

OleksandrMoroz24:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+from app.errors import (
```

app/main.py
```diff
@@ -1 +1,19 @@
...
+    masks_to_buy = sum(
+        1 for friend in friends if not friend.get("wearing_a_mask", False)
+    )
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError as e:
+            return str(e)
+        except NotWearingMaskError:
+            pass
```
Y-Havryliv, line 15: using pass in exceptions is not the best practice
https://stackoverflow.com/questions/21553327/why-is-except-pass-a-bad-programming-practice

Y-Havryliv (changes requested):

app/cafe.py
OleksandrMoroz24: like this?
```
from app.errors import (

    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```

Y-Havryliv: no, like this
```python
import datetime

from app.errors import (
```

Y-Havryliv (approved):
gj!

PR #561, opened 2023-08-22
Solution

VasylTurok:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Cafe:
+
+    def __init__(self, name: str) -> None:
+
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+
+        if "vaccine" not in visitor:
```
pavlejviki, line 14: use get() method when checking the presence or accessing the value of a key in all your if statements

app/errors.py
```diff
@@ -0,0 +1,17 @@
+class VaccineError(Exception):
+    pass
+
+
+class NotVaccinatedError(VaccineError):
+    def __str__(self) -> str:
+        return "Not vaccinated"
```
pavlejviki, line 7: your error messages should be more descriptive

app/main.py
```diff
@@ -1 +1,18 @@
...
+from app.errors import (NotVaccinatedError,
+                        NotWearingMaskError, OutdatedVaccineError)
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
pavlejviki, line 11: You can catch the parent class VaccineError instead of the two child classes

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (NotVaccinatedError,
```
pavlejviki, line 2: import is not formatted properly

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #562, opened 2023-08-23
Solution

ivanzh13:


ArsenPidhoretskyi (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None:
+        vaccine = visitor.get("vaccine")
+        if not vaccine:
+            raise NotVaccinatedError("Visitor does't have vaccine.")
+        if date.today() > vaccine["expiration_date"]:
```
vasylhnatiuk, line 17: use get() here also

vasylhnatiuk (changes requested):

app/cafe.py
ivanzh13: If I get None, I still get an error

ivanzh13: Do you can explain me, why use get here

vasylhnatiuk: Here is more about style of code: if you use get() in one place, use it in all places. And you will not have KeyError

vasylhnatiuk (approved):

PR #567, opened 2023-08-26
Develop

VASO304:


Kroll410 (changes requested):
add descriptive error message

VASO304:
> add descriptive error message

Hi, Dmytro! Unfortunately, it is unclear to me what kind of description is needed. There are comments on the errors in the error file. Could you please advise me?

PR #570, opened 2023-08-28
'Solution'

o-sapohova:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Everyone must be vaccinated")
+        if date.today() > visitor["vaccine"].get("expiration_date"):
+            raise OutdatedVaccineError("The vaccine must not be expired")
+        if visitor["wearing_a_mask"] is False or\
+                "wearing_a_mask" not in visitor:
```
EdAlekseiev, line 16: Use `()` instead of `/` for separate line

EdAlekseiev, line 16:
```suggestion
        if ("wearing_a_mask" not in visitor
             or visitor["wearing_a_mask"] is False):

```

EdAlekseiev (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,18 @@
+from app.errors import NotVaccinatedError,\
+    OutdatedVaccineError, NotWearingMaskError
```
olena-hul, line 2: Use `()` instead of `\`

olena-hul (changes requested):

olena-hul (changes requested):
Fix comment

olena-hul (approved):
Good job!

PR #573, opened 2023-09-01
Solution by IvDyriv

IvDyriv:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #580, opened 2023-09-09
solution

d-beschastnyi:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError
+        elif not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 16: Use .get because wearing_a_mask may not exist into visitor

ArsenPidhoretskyi (changes requested):

app/cafe.py
ArsenPidhoretskyi: Not fixed

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #590, opened 2023-09-22
Solution

SerhiiBudko:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
```
viktoria-rybenchuk, line 1: add a blank line between imports according to pep8  https://peps.python.org/pep-0008/#imports

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #591, opened 2023-09-24
'Solution'

nsdNite:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None | str:
+        current_date = date.today()
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("You have no vaccine certificate."
+                                     " You shall not pass!")
+        if visitor["vaccine"]["expiration_date"] < current_date:
```
Dimosphen1, line 18: You should use this way of retrieving values from dict only in case you are sure of the structure of the incoming dict. In production settings, it would be better to use `.get()` in case there is room for a potential change of the dict structure.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #598, opened 2023-09-29
Implemented lockdown cafe task

sberdianskyi:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("You have to be vaccinated.")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Your vaccine is expired.")
+        if visitor["wearing_a_mask"] is False:
```
ArsenPidhoretskyi, line 16: Use .get because wearing_a_mask may not exist into visitor

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #599, opened 2023-09-30
Solution

valikdeb:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```

Bohdan-Salamakha (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("NotVaccinatedError")
```
viktoria-rybenchuk, line 13: Use descriptive error messages. Fix in all places

app/main.py
```diff
@@ -1 +1,18 @@
...
+    without_mask_count = 0
+    for person in friends:
+        try:
+            cafe.visit_cafe(person)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            without_mask_count += 1
+
+    if without_mask_count != 0:
```

viktoria-rybenchuk (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
+import datetime
+from app.errors import (OutdatedVaccineError,
+                        NotVaccinatedError,
+                        NotWearingMaskError)
```
viktoria-rybenchuk, line 4: fix order imports related to pep8 https://peps.python.org/pep-0008/#imports

viktoria-rybenchuk (changes requested):
Fix flake8

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
```
pavlejviki, line 15: using keys() is redundant, as `in `anyway checks the keys of the dictionary.

but better use .get() method when accessing the value of a key

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #600, opened 2023-09-30
solution

Depaton:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
```
pavlejviki, line 12: use the .get() method when retrieving the value of a key to ensure it's present and not None.

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+            raise NotVaccinatedError("Visitor is not vaccinated.")
+        if visitor["vaccine"]["expiration_date"] \
```
pavlejviki, line 14: avoid using \, use parenthesis instead

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
pavlejviki, line 4: you could use a single line to import multiple classes from the same module

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError
+from app.errors import NotWearingMaskError
```
pavlejviki, line 3: you could use a single line to import multiple classes from the same module

pavlejviki (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        vaccine = visitor.get("vaccine")
+        if vaccine is None:
+            raise NotVaccinatedError("Visitor is not vaccinated.")
+        if vaccine.get("expiration_date") < datetime.datetime.now().date():
+            raise OutdatedVaccineError("Vaccine has expired")
+        if visitor["wearing_a_mask"] is False:
```
viktoria-rybenchuk, line 19: you don't need to compare a boolean value to False explicitly

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #602, opened 2023-10-03
Cafe_sollution

Filllick:


app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor isn't vaccinated")
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Visitor's vaccine has expired")
+
+        if not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 20: wearing_a_mask may not exist, in this case use .get

ArsenPidhoretskyi (changes requested):

Filllick:
Used .get to check if wearing_a_mask in visitor dictionary

vasylhnatiuk (approved):

PR #605, opened 2023-10-07
Solution

loplicat:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from datetime import date
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError)
```
viktoria-rybenchuk, line 5:
```suggestion
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
    )
```

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from datetime import date
```
viktoria-rybenchuk, line 1: fix imports ordering according to PEP8  https://peps.python.org/pep-0008/#imports

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("Vaccine is expired")
+        if visitor["wearing_a_mask"] is False:
```
viktoria-rybenchuk, line 17: Check if the visitor is not wearing a mask

viktoria-rybenchuk (changes requested):

app/cafe.py
loplicat: Thanks for the comments, I already implemented these changes :)

viktoria-rybenchuk (approved):
Well done!

PR #607, opened 2023-10-12
Solution

YuliaHladyshkevych:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+class Cafe:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor don`t have a vaccine.")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("The vaccine must not be expired!")
```
Familenko, line 18: It is better to create the same variable instead of accessing the 3rd nesting level

LiudmylaKulinchenko (approved):
Good job! 🚀

PR #608, opened 2023-10-12
solution

Stainback:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" in visitor:
+            if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
EdAlekseiev, line 12: Use `get` method for getting value by key to avoid KeyError

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+                raise OutdatedVaccineError("Vaccine is expired")
+        else:
+            raise NotVaccinatedError("Vaccine is required")
+
+        if not visitor["wearing_a_mask"]:
```
EdAlekseiev, line 17: The same comment as above

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,27 @@
...
+    for visitor in visitors:
+        try:
+            cafe.visit_cafe(visitor)
+        except VaccineError:
+            violations = -1
+            break
+        except NotWearingMaskError:
+            violations += 1
+
+    return violations
```
ArsenPidhoretskyi, line 27: This is so strange, please remove this function into go_to_cafe

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

app/main.py
Stainback: Tried to make functions more atomic by separating error handling from message output. But I admit that logic might look distracting.

PR #610, opened 2023-10-13
sol

Familenko:


app/cafe.py
```diff
@@ -0,0 +1,29 @@
...
+        vaccine_exp = vaccinated["expiration_date"] if vaccinated else None
+
+        if vaccinated is None:
+            raise NotVaccinatedError("Beware, conspiracy theorist")
+
+        if vaccine_exp < current_date:
+            raise OutdatedVaccineError("Person without plans")
+
+        if not mask_on_face:
+            raise NotWearingMaskError("Big nose person!")
```
Naz-iv, line 27: good idea for error message, not sure if it will be counted as "descriptive error messages"

app/cafe.py
```diff
@@ -0,0 +1,29 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        vaccinated = visitor.get("vaccine")
+        current_date = datetime.date.today()
+        mask_on_face = visitor["wearing_a_mask"]
+        vaccine_exp = vaccinated["expiration_date"] if vaccinated else None
```
Naz-iv, line 18: maybe you could avoid creating additional variables by putting information in  if statement directly for example || if not visitor.get("vaccine") ||

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+from app.errors import (
+    VaccineError,
+    NotWearingMaskError
+)
+
```
Naz-iv, line 5: Not 100 % sure, but I think according to conventions you should separate imports with blank lines if they are from different blocks (Standard Library Imports, Third-Party Library Imports, Local Imports). Both errors and cafe are local imports, so maybe no need to blank line

Naz-iv (approved):
good job

app/cafe.py
Familenko: depends on teamlead desicion))

app/main.py
Familenko: yeap, agree

fsocie7y (approved):
well done ^_^

PR #611, opened 2023-10-13
Develop

Naz-iv:


YuliaHladyshkevych (approved):

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Cafe:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("Vaccination certificate missing!")
+        if visitor["vaccine"]["expiration_date"] < date.today():
```
LiudmylaKulinchenko, line 16: It looks like u did't use `get` here for some purpose but I don't get it
Use same code style

app/main.py
```diff
@@ -1 +1,16 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    return (f"Friends should buy {masks_to_buy} masks"
+            if masks_to_buy
+            else f"Friends can go to {cafe.name}")
```
LiudmylaKulinchenko, line 16: avoid using ternary operator here to increase readability

app/cafe.py
Naz-iv: corrected

app/main.py
Naz-iv: corrected

LiudmylaKulinchenko (approved):
Great! 🔥

PR #614, opened 2023-10-14
Solution

prochigor:


petrykivd (approved):

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if visitor.get("vaccine"):
+            expiration_date = visitor.get("vaccine").get("expiration_date")
+        else:
+            raise app.errors.NotVaccinatedError
```
LiudmylaKulinchenko, line 14: try to rewrite to increase readability

app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+
+import app.errors
```
LiudmylaKulinchenko, line 3: import just needed classes

LiudmylaKulinchenko (changes requested):

app/cafe.py
prochigor: Okay, I created new variables for calling dict

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🌷

PR #615, opened 2023-10-14
Solution

petrykivd:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+import datetime
```
prochigor, line 6: change import order, first build-in, after your imports with an empty line between them

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("All friends should be vaccinated.")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("The vaccine must not be expired!")
```
prochigor, line 17: you can init these messages in VactineError class with __str__ method like "All friends should be vaccinated"
image
or inside child classes your massages, but it **up to you**

prochigor (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🚀

PR #617, opened 2023-10-14
Solution for py_lockdown_cafe

wataschi:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
```
viktoria-rybenchuk, line 14:
```suggestion
         if not visitor.get("vaccine"):
```

viktoria-rybenchuk (changes requested):

wataschi:
@viktoria-rybenchuk

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError
+        if "wearing_a_mask" in visitor and visitor["wearing_a_mask"] is False:
```
viktoria-rybenchuk, line 18: Use the `get` method for dictionary access can be a better approach. It's a good practice for handling cases where the presence of a key is uncertain.

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #619, opened 2023-10-15
Lockdown Cafe solution

Lyutillis:


RVChornyy (approved):

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+                     OutdatedVaccineError,
+                     NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine", None):
```
LiudmylaKulinchenko, line 13: You can use `if "vaccine" not in visitor` to check keys here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great job! 🐣

PR #622, opened 2023-10-15
Solution

mykolamateichuk:


app/main.py
```diff
@@ -1 +1,22 @@
...
+    OutdatedVaccineError
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
Lyutillis, line 14: Here you can use VaccineError to catch both errors

Lyutillis (approved):

app/main.py
LiudmylaKulinchenko: Fix that

app/main.py
```diff
@@ -1 +1,22 @@
...
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy != 0:
```
LiudmylaKulinchenko, line 19: it is okay, but u can just if masks_to_buy: , since it will return True only if masks_to_buy > 0

LiudmylaKulinchenko (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError(f"{visitor['name']} is not vaccinated!")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError(f"{visitor['name']}'s"
+                                       f" vaccine is expired!")
+        if not visitor.get("wearing_a_mask") or not visitor["wearing_a_mask"]:
```
LiudmylaKulinchenko, line 19: You duplicating conditions here, remove one

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError(f"{visitor['name']} is not vaccinated!")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError(f"{visitor['name']}'s"
+                                       f" vaccine is expired!")
```
LiudmylaKulinchenko, line 18: I'm not sure if it's fine to write the name because you don't know if there just one friend hat doesn't wear a mask

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+    OutdatedVaccineError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
```
LiudmylaKulinchenko, line 14: You can use `if "vaccine" in visitor` here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 😸

PR #625, opened 2023-10-16
KFC on lockdown

SelFisher1488:


app/cafe.py
```diff
@@ -0,0 +1,17 @@
+import datetime
+from app.errors \
```
Shatrovskyi, line 2: don't use \
Instead, you can use (NotVaccinatedError, 
                               OutdatedVaccineError, 
                               NotWearingMaskError)

app/main.py
```diff
@@ -1 +1,18 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
Shatrovskyi, line 16: It's not necessary to check >. you can just do this:

If masks_to_buy:
     print("-------------------")

Shatrovskyi (approved):

app/cafe.py
fsocie7y: fix it

app/main.py
fsocie7y: fix it also

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #626, opened 2023-10-16
Solution

katyakoziar:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if visitor.get("vaccine") is None:
+            raise NotVaccinatedError("Visitor is not vaccinated.")
+        elif visitor["vaccine"].get("expiration_date") < datetime.date.today():
```
SevKrok, line 17: you not need 'elif' here couse func will end when raise error

SevKrok (approved):
Well done ;)

app/cafe.py
LiudmylaKulinchenko: fix that

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if visitor.get("vaccine") is None:
```
LiudmylaKulinchenko, line 15: you can check if there is a key like this `if "vaccine" in visitor`

app/main.py
```diff
@@ -1 +1,16 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy > 0:
```
LiudmylaKulinchenko, line 14: it is okay, but u can just if masks_to_buy: , since it will return True only if masks_to_buy > 0

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great! 🚀

PR #627, opened 2023-10-16
Develop

Andru37:


Andru37:
In this task is not emphasized that it is necessary to count people without a mask

EdAlekseiev (approved):
Good job

PR #632, opened 2023-10-16
Solution

D1st3f:


fedorov-s-od (approved):
The solution looks good. But it would probably be better to put descriptive messages for all types of errors in their str methods )

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("All friends should be vaccinated")
```
LiudmylaKulinchenko, line 18: use descriptive message here. They have vaccines, but some of them have them expired, fix that

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("Friends should buy masks")
```
LiudmylaKulinchenko, line 20: fix message, now it looks like to visit a cafe everyone should buy a mask, but they should just wear them

LiudmylaKulinchenko (changes requested):

app/cafe.py
D1st3f: fixed

app/cafe.py
D1st3f: if i change text, pytest will give an error

LiudmylaKulinchenko (approved):
Well done! 🚀

PR #634, opened 2023-10-16
Implement py-lockdown-cafe

SevKrok:


Katherine-Greg (approved):
Good job! 👍

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
LiudmylaKulinchenko, line 16: it is okay, but u can just if masks_to_buy: , since it will return True only if masks_to_buy > 0

LiudmylaKulinchenko (approved):
Great! 🚀

PR #635, opened 2023-10-16
'Solution'

fedorov-s-od:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
+from datetime import date
+from app.errors import (NotVaccinatedError,
+                        NotWearingMaskError,
+                        OutdatedVaccineError)
+
```
Wonsky1, line 5: Maybe it would be better to implement import fields separately

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError()
+
+        expiration_date = visitor["vaccine"].get("expiration_date")
+        if expiration_date < date.today():
+            raise OutdatedVaccineError()
+
+        if not visitor.get("wearing_a_mask"):
+            raise NotWearingMaskError()
```
Wonsky1, line 20: Maybe it would be better to add default value for get

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
```
Wonsky1, line 2: To my mind same as in cafe.py, make separately

Wonsky1 (approved):

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+                        OutdatedVaccineError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError()
```
LiudmylaKulinchenko, line 13: You can remove parentheses

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #637, opened 2023-10-16
Solution

skoval7:


app/main.py
```diff
@@ -1 +1,22 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except NotVaccinatedError:
+            return "All friends should be vaccinated"
+        except OutdatedVaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
LiudmylaKulinchenko, line 19: it is okay, but u can just if masks_to_buy: , since it will return True only if masks_to_buy > 0

LiudmylaKulinchenko (approved):
Well done! 🔥

PR #639, opened 2023-10-17
Solution Lockdown cafe

Oxbay:


oleksandr-pupchenko (approved):

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("You must be vaccinate")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("Your vaccine is outdated")
+        if ("wearing_a_mask" not in visitor
+                or visitor["wearing_a_mask"] is False):
```
fsocie7y, line 18: u can use .get() instead of this long one

fsocie7y (changes requested):

app/cafe.py
fsocie7y: fix that)

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #641, opened 2023-10-18
Solution

Andrii-Silenko:


app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+        visitor_name = visitor.get("name")
+        if visitor.get("vaccine") is None:
+            raise NotVaccinatedError(f"{visitor_name} "
+                                     f"should vaccine to visit this cafe")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError(f"{visitor_name}'s "
+                                       f"vaccine is expired")
+        elif visitor.get("wearing_a_mask") is not True:
```
fsocie7y, line 19: u can use `not visitor.get("wearing_a_mask")` insteaf of `is not True`

fsocie7y (changes requested):

LiudmylaKulinchenko (approved):
Well done 😋

PR #643, opened 2023-10-19
Solution

kravetb:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> Any:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("All friends should be vaccinated")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
EdAlekseiev, line 16: Use method `get` of dict to avoid `KeyError` exception

EdAlekseiev, line 16:
```suggestion
        elif visitor["vaccine"].get("expiration_date") < datetime.date.today():

```

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+            raise OutdatedVaccineError("All friends should be vaccinated")
+        elif not visitor["wearing_a_mask"]:
```

app/main.py
```diff
@@ -1 +1,21 @@
...
+from typing import Any
+from app.cafe import Cafe
+from app.errors import (
+    VaccineError,
+    NotWearingMaskError)
+
+
+def go_to_cafe(friends: list, cafe: "Cafe") -> Any:
+    try:
+        count = 0
```
EdAlekseiev, line 10: `count` is not the clearer name. Change it, please.

EdAlekseiev (changes requested):

PR #647, opened 2023-10-23
Solution

okien1:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(
+            self,
+            visitor: dict
+    ) -> str | NotVaccinatedError | OutdatedVaccineError | NotWearingMaskError:
```
pavlejviki, line 17: The exceptions are not part of the return type, so they shouldn't be included in the type hint for the return value.

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("Person is not vaccinated!")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
pavlejviki, line 20: use .get() when accessing a key

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #653, opened 2023-10-26
solution_py-lockdown-cafe_1

VladKrit-commits:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from typing import List
```
Dimosphen1, line 1: Consider separating different groups of imports by an empty string, here you use two groups - built-in package import and import of local files.

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
Dimosphen1, line 18: In the production environment, prefer using .get() to obtain values from dicts in case you ar uncertain in the incoming dict structure.

Dimosphen1 (approved):
LGTM, well done! Several comments were left for consideration.

PR #654, opened 2023-10-26
Solution COVID-19

hbilous:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitors must be vaccinated")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
EdAlekseiev, line 14: Use method get of dict to avoid KeyError exception

EdAlekseiev (changes requested):

YukihiroSM (approved):
Good job!

PR #657, opened 2023-10-29
Solution

makukh:


app/main.py
```diff
@@ -1 +1,31 @@
...
+        if vaccinated_friends == len(friends):
+            if mask_wearing_friends == len(friends):
+                return f"Friends can go to {cafe.name}"
+            else:
+                masks_to_buy = len(friends) - mask_wearing_friends
+                return f"Friends should buy {masks_to_buy} masks"
+        else:
+            raise VaccineError("All friends should be vaccinated")
+    except VaccineError as e:
+        return str(e)
```
ArsenPidhoretskyi, line 31: Your code so complicated. Try to simplify this. Use 1 forloop and inside try-except statements

ArsenPidhoretskyi (changes requested):

makukh (commented):
Refactored code with one loop in main instead of two

YukihiroSM (approved):
Looks good!

PR #659, opened 2023-11-03
Solution

CrazyHilly:


app/cafe.py
NikOneZ1, line 1: It would be better to use .get() method instead of the [] to avoid KeyError and simplify the code

NikOneZ1 (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+            raise errors.NotVaccinatedError(
+                f"{visitor.get('name')} is not vaccinated"
+            )
+
+        if date.today() > visitor["vaccine"]["expiration_date"]:
+            raise errors.OutdatedVaccineError(
+                f"Vaccine for {visitor.get('name')} has expired"
+            )
+
+        if not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 21:
```suggestion
        if not visitor.get("wearing_a_mask"):

```
wearing_a_mask may not exist into visitor

ArsenPidhoretskyi (changes requested):

NikOneZ1 (approved):
Good job!

PR #660, opened 2023-11-04
Implemented py-lockdown-cafe

Aeros72:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+from datetime import date
+from typing import Union
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
```
viktoria-rybenchuk, line 5: Imports should be grouped in the following order:

Standard library imports.
Related third party imports.

viktoria-rybenchuk, line 5: Use ( ) while importing multiple classes/modules, follow example formatting imports in checklist

app/main.py
```diff
@@ -1 +1,18 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy == 0:
```

app/main.py
```diff
@@ -1 +1,18 @@
...
+        return f"Friends can go to {cafe.name}"
+    else:
```
viktoria-rybenchuk, line 17: else is redundant

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (VaccineError,
+                        NotWearingMaskError)
```
viktoria-rybenchuk, line 3: fix imports according to checklist

viktoria-rybenchuk (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+    def visit_cafe(self, visitor: dict) -> Union[str, Exception]:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("Visitor should be vaccinated")
+        else:
+            if visitor["vaccine"]["expiration_date"] < date.today():
+                raise OutdatedVaccineError("Vaccine should be up-to-date")
+            else:
+                if visitor["wearing_a_mask"] is False:
+                    raise NotWearingMaskError("Visitor should wear a mask")
+        return f"Welcome to {self.name}"
```
Dimosphen1, line 24: Write the `visit_cafe` method without using nesting if conditions and `else` statements. Take into account that the code will not execute anyway if any exception is raised.

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> Union[str, Exception]:
```
Dimosphen1, line 15: Do not indicate raised `Exception` in return type annotation, there is [no need for it](https://peps.python.org/pep-0484/#exceptions). Leave just `str`.

Dimosphen1 (changes requested):

vsmutok (approved):
Looks Good 👍

PR #661, opened 2023-11-05
need help

Andrey1104:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(
+            self,
+            visitor: dict
+    ) -> str | VaccineError | OutdatedVaccineError:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("visitor is not vaccinated")
+        if datetime.date.today() > visitor["vaccine"]["expiration_date"]:
```
Dimosphen1, line 21: It is good practice to place in the first position for comparison an incoming value and at the second position a value it is compared against. Change the position of arguments that you compare on this line.

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(
+            self,
+            visitor: dict
+    ) -> str | VaccineError | OutdatedVaccineError:
```
Dimosphen1, line 18: You do not need to indicate exceptions in return type annotation, they are out of scope [for this feature](https://peps.python.org/pep-0484/#exceptions).

app/cafe.py
```diff
@@ -0,0 +1,25 @@
+import datetime
+
+from app.errors import (
+    VaccineError,
```
Dimosphen1, line 4: You do not need to import `VaccineError` here.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import NotWearingMaskError, VaccineError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
Dimosphen1, line 5: You could be more specific in the future indicating a list of what entities is expected for the variable.

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("visitor is not vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
Dimosphen1, line 17: Whenever you are not sure about the incoming dict structure, prefer using `.get()` to retrieve values.

Dimosphen1 (approved):
LGTM, well done! Several comments were left for consideration.

PR #663, opened 2023-11-05
Solution

AnyoneClown:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+from app.errors import (
```
Dimosphen1, line 2: Do not forget to separate different groups of imports with a blank line according to PEP-8.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #669, opened 2023-11-09
'Solution'

yul1valko:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("All friends should be vaccinated")
+
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("All friends should be vaccinated")
+
+        if not visitor["wearing_a_mask"]:
```
viktoria-rybenchuk, line 22: Using `get` with a default value ensures a safe check for the presence of the key "wearing_a_mask" in the dictionary, avoiding a `KeyError` and providing a default behavior when the key is absent.

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+            raise NotWearingMaskError(" ")
```
viktoria-rybenchuk, line 23: Add descriptive error message

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from app.cafe import Cafe
+
```
viktoria-rybenchuk, line 2: remove blank line. You should put a blank line just  between different group of imports. Read PEP8 https://peps.python.org/pep-0008/#imports

app/main.py
```diff
@@ -1 +1,21 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+
+        except NotWearingMaskError:
+            amount_of_masks += 1
+
+    if amount_of_masks > 0:
```
viktoria-rybenchuk, line 18: Don't compare a variable to 0 if there is no need. Read checklist

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #672, opened 2023-11-12
Solution

sixsixsixsamurai:


app/main.py
```diff
@@ -1 +1,33 @@
...
+                raise OutdatedVaccineError
+            elif friend["vaccine"]["expiration_date"] < datetime.date.today():
+                raise OutdatedVaccineError
+
+        for friend in friends:
+            if friend["wearing_a_mask"] is False:
+                friend_with_no_mask += 1
+
+        if friend_with_no_mask:
+            raise NotWearingMaskError(friend_with_no_mask)
```
Dimosphen1, line 24: There is no need to perform these checks here, relay only on the `visit_cafe` method, it will raise necessary errors. Remove the code from lines 13 to 24 to not violate the DRY.

app/errors.py
```diff
@@ -0,0 +1,19 @@
...
+class OutdatedVaccineError(VaccineError):
+    pass
+
+
+class NotWearingMaskError(Exception):
+    def __init__(self, friend_with_no_mask: int = 0) -> None:
+        self.friend_with_no_mask = friend_with_no_mask
+
+    def __str__(self) -> str:
+        return f"Friends should buy {self.friend_with_no_mask} masks"
```
Dimosphen1, line 19: There is no need to implement this logic, place the required logic in the `go_to_cafe` and replace the code with `pass`.

app/errors.py
```diff
@@ -0,0 +1,19 @@
+class VaccineError(Exception):
+    def __str__(self) -> str:
+        return "All friends should be vaccinated"
```
Dimosphen1, line 3: It would be better to return this specific string in the required place directly, because `VaccineError` can apply to not only friends but to other people or animals as well out of the current context. Besides that, all inherited exceptions would have this text, and the vaccine absence can be not the only reason for this error. Replace this code with `pass`.

app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+import datetime
+from app.cafe import Cafe
```
Dimosphen1, line 2: Separate your local imports and built-in library imports with one blank line according to PEP-8.

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+        if friend_with_no_mask:
+            raise NotWearingMaskError(friend_with_no_mask)
+
+        for friend in friends:
+            cafe.visit_cafe(visitor=friend)
+    except VaccineError as e:
+        return str(e)
+    except NotWearingMaskError as e:
+        e.friend_with_no_mask = friend_with_no_mask
```
Dimosphen1, line 31: You could increment the `friend_with_no_mask` with 1 here and, after the loop execution, return the required text in case friends without masks are present.

app/cafe.py
```diff
@@ -0,0 +1,25 @@
+import datetime
+from app.errors import (
```
Dimosphen1, line 2: Separate your local imports and built-in library imports with one blank line according to PEP-8.

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        current_date = datetime.datetime.today().date()
+
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError
+
+        if visitor["vaccine"]["expiration_date"] < current_date:
```
Dimosphen1, line 19: Prefer using `get()` when you are not sure about the incoming dict structure.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,21 @@
...
+    VaccineError
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    friend_with_no_mask = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(visitor=friend)
+        except (VaccineError):
```
Dimosphen1, line 14: There is no need to place single exception in `()`, leave it as follows in the future
```suggestion
        except VaccineError:
```

Dimosphen1 (approved):

PR #674, opened 2023-11-14
Solution

dkibalenko:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("Visitor should be vaccinated.")
+        if not visitor["vaccine"]["expiration_date"] >= datetime.date.today():
```
pavlejviki, line 16: better to use get() in all places where you access a key

pavlejviki (changes requested):

app/cafe.py
dkibalenko: > better to use get() in all places where you access a key

exactly, thanks!

Polyakiv-Andrey (approved):
Good job!

dkibalenko:
Thanks!

PR #676, opened 2023-11-16
Solution

mrve4nick:


Polyakiv-Andrey (changes requested):
Test not passed!

mrve4nick:
I keep getting errors 
Screenshot from 2023-11-16 21-51-56
Screenshot from 2023-11-16 21-52-19
Somehow in some tests OutdatedVaccineError is not raising and raised NotWearingMaskError instead

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import app.errors
+import datetime
```
Dimosphen1, line 2: Order your imports according to PEP-8. Besides that, leave blank lines between different groups of imports. This comment relates to `app/main.py` as well.

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import app.errors
```
Dimosphen1, line 1: Import the exact error classes from `app.errors` that you will need to call, it is a good practice to import classes instead of calling them from the module in this case. This comment relates to `app/main.py` as well.

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | None:
+        if "vaccine" not in visitor:
+            raise app.errors.NotVaccinatedError("Visitor is not vaccinated")
+
+        if visitor["vaccine"]["expiration_date"] <= datetime.date.today():
```
Dimosphen1, line 13: Change this condition, the `expiration_date` should be less than `datetime.date.today()`, not less than equal.

app/main.py
```diff
@@ -1 +1,15 @@
...
+    try:
+        for friend in friends:
+            cafe.visit_cafe(friend)
+
+        return f"Friends can go to {cafe.name}"
+    except app.errors.VaccineError:
+        return "All friends should be vaccinated"
+    except app.errors.NotWearingMaskError:
+        masks_to_buy = sum(1 for friend in friends if not friend.get("wearing_a_mask", False))
+        return f"Friends should buy {masks_to_buy} masks"
```
Dimosphen1, line 15: Change your error processing, you should create a try except block inside of the for loop, because `NotWearingMaskError` can raise earlier than `VaccineError`, but the presence of the `VaccineError` leads to the refusal in visiting cafe to all friends despite the presence or absence of masks.

Dimosphen1 (changes requested):
Several changes were requested. All tests should pass upon the next review request.

Polyakiv-Andrey (changes requested):
Tests not passed!

PR #678, opened 2023-11-20
Lockdown-Cafe

MykolaMazh:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Everyone should be vaccinated")
+        else:
```
pavlejviki, line 14: else is redundant

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
```
pavlejviki, line 12: use get() when accessing a key. fix in all places

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+            raise NotVaccinatedError("Everyone should be vaccinated")
+        else:
+            if datetime.date.today() > visitor["vaccine"]["expiration_date"]:
+                raise OutdatedVaccineError("Vaccination should be valid")
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError(
+                "Even vaccinated visitors should were masks")
+        else:
```
pavlejviki, line 20: else is redundant

pavlejviki (changes requested):

app/cafe.py
MykolaMazh:
```python
if "vaccine" not in visitor:
```
here i just check the existence of the vaccine key
"If the visitor does not have a vaccine key, it means that he is not vaccinated. In this case, the method must raise a NotVaccinatedError exception."

why? 
```python
visitor.get("vaccine")
                .get("expiration_date")
```
is better than
```python
visitor["vaccine"]["expiration_date"]
```
cause we know for sure the vaccine key is

with 
```python
visitor.get("wearing_a_mask")
```
i fully agree

```python
kfc = Cafe("KFC")
visitor = {
    "name": "Paul",
    "age": 23,
}
kfc.visit_cafe(visitor)     # NotVaccinatedError
```
```python
import datetime

kfc = Cafe("KFC")
visitor = {
    "name": "Paul",
    "age": 23,
    "vaccine": {
        "expiration_date": datetime.date(year=2019, month=2, day=23)
    }
}
kfc.visit_cafe(visitor)     # OutdatedVaccineError
```
here i see a visitor without wearing_a_mask key can exist

pavlejviki: you are right `if "vaccine" not in visitor:` is fine, if you prefer like that

but you need to check if "expiration_date" exists in visitor["vaccine"]

pavlejviki (commented):
Well done!

pavlejviki (approved):

PR #680, opened 2023-11-21
Solution

maxvell4:


app/cafe.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        current_date = datetime.date.today()
+
+        if "vaccine" in visitor:
+            expiration_date = visitor["vaccine"]["expiration_date"]
+            if expiration_date < current_date:
+                raise OutdatedVaccineError("Expiration date out of time")
```
Dimosphen1, line 20: You could avoid using two nested if conditions in case you will use dict chaining with `get()` and using dict as a default return value, e.g.:
```suggestion
        if visitor.get("vaccine", {}).get("expiration_date", current_date) <= current_date:
            raise OutdatedVaccineError("Expiration date out of time")
```
You could modify the logic to satisfy the tests but keep the overall structure mentioned above.

app/main.py
```diff
@@ -1 +1,24 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    errors = []
+    masks_to_buy = 0
+
+    for visitor in friends:
+
+        try:
+            cafe.visit_cafe(visitor)
+        except VaccineError as e:
+            errors.append(e)
```
Dimosphen1, line 16: There is no need to allocate `errors` variable and append an exception to it because in case the `VaccineError` will be raised, the next step would be a `return` statement without further function execution. Remove this line, the `errors` variable, and other logic related to it.

app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+
+from app.cafe import Cafe
+
+from app.errors import (VaccineError, NotWearingMaskError)
```
Dimosphen1, line 4: Place each Exception on a new line as it is done in `app/cafe.py` in case you wrap them in parenthesis.

Dimosphen1 (changes requested):

app/main.py
maxvell4: i don`t understand

Polyakiv-Andrey (approved):
Good job!

PR #682, opened 2023-11-21
Solution

OleksandrMV:


YukihiroSM (changes requested):
Add some code, please.
For future: you can create a pr after you pushed all required commits. 
Also, please, try to not close PRs, it may raise an issue with Mate services.

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("NotVaccinatedError")
+        elif visitor["vaccine"]["expiration_date"] < date.today():
```
EdAlekseiev, line 13: Use method get of dict to avoid KeyError

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        elif not visitor["wearing_a_mask"]:
```
EdAlekseiev, line 15: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #683, opened 2023-11-22
'Solution'

SerhiiDomik:


app/main.py
```diff
@@ -1 +1,21 @@
...
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    not_vaccinated = 0
+
+    for visitor in friends:
+        try:
+            cafe.visit_cafe(visitor)
+        except VaccineError:
+            not_vaccinated += 1
```
ArsenPidhoretskyi, line 13:
```suggestion
            return "All friends should be vaccinated"
```
Really, we do not need calculate this

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
```suggestion
def go_to_cafe(friends: list[dict], cafe: Cafe) -> str: 
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #684, opened 2023-11-22
Solution

Liliyakhu:


app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str | Exception:
```
Dimosphen1, line 5: Do not use raised `Exception` in the return type annotation, it is not supposed to be present according [to the documentation](https://peps.python.org/pep-0484/#exceptions). Remove `Exception` in the return type annotation.

app/main.py
```diff
@@ -1 +1,23 @@
...
+    masks_to_buy = 0
+    quantity_of_friends = 0
```
Dimosphen1, line 7: There is no need for the `quantity_of_friends` variable and logic related to it. Remove it from your code.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,20 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
viktoria-rybenchuk, line 17: Don't compare a variable to 0 if there is no need. Read checklist

viktoria-rybenchuk (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,28 @@
...
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(
+                f"The {self.name} visitor "
+                f"{visitor['name']} doesn't have a vaccine!"
+            )
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError(
+                f"The {visitor['name']}'s vaccine has expired!"
+            )
+        if visitor["wearing_a_mask"] is False:
```
pavlejviki, line 24: better use get() method when accessing a key in a dictionary

pavlejviki (changes requested):

vsmutok (approved):
Good Job 👍

PR #685, opened 2023-11-23
Solution for py_lockdown_cafe

VladyslavBazhyn:


app/main.py
```diff
@@ -1 +1,31 @@
...
+            pass
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+
+        except NotWearingMaskError as n:
+            masks_to_buy = len([friend for friend in friends
+                                if friend.get("wearing_a_mask") is False])
+            return str(n) + f"{masks_to_buy} masks"
```
EdAlekseiev, line 29: Don't call `visit_cafe` twice. You can catch all exception inside one loop

EdAlekseiev (changes requested):

app/main.py
VladyslavBazhyn: I can, but in this case I will take only first Error which will appearing. In some tests it's important that even if appearing both NotWearingMaskError and VaccineError only VaccineError will raise. Even if NotWearingMaskError appears first. To solve this I made two independent loops. May be you can tell how to make this prioritet ominside one loop?

VladyslavBazhyn:
How to make rpioritet in errors without two loops?

app/cafe.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
```
EdAlekseiev, line 17: You must raise an object of error.
```suggestion
            raise NotVaccinatedError()

```

app/cafe.py
```diff
@@ -0,0 +1,29 @@
...
+
+        visitor_vaccine_date = (
+            visitor.get("vaccine").get("expiration_date")
+        )
+
+        if visitor_vaccine_date < day_today:
+            raise OutdatedVaccineError
```
EdAlekseiev, line 24:
```suggestion
            raise OutdatedVaccineError()

```

app/cafe.py
```diff
@@ -0,0 +1,29 @@
...
+
+        if visitor.get("wearing_a_mask") is False:
+            raise NotWearingMaskError
```
EdAlekseiev, line 27:
```suggestion
            raise NotWearingMaskError()

```

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+day_today = datetime.date.today()
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    vaccine_error_encountered = False
+    not_wearing_mask_error = None
```
EdAlekseiev, line 11: You don't need this variables

app/main.py
```diff
@@ -1 +1,30 @@
...
+    masks_to_buy = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError as v:
+            vaccine_error_encountered = True
+            return str(v)
```
EdAlekseiev, line 19:
```suggestion
            return str(v)
```

app/main.py
```diff
@@ -1 +1,30 @@
...
+        except NotWearingMaskError as n:
+            if not vaccine_error_encountered:
+                masks_to_buy += 1
+                not_wearing_mask_error = n
```
EdAlekseiev, line 23:
```suggestion
            masks_to_buy += 1
            not_wearing_mask_error = n
```

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+    if vaccine_error_encountered:
+        return str(vaccine_error_encountered)
+    elif not_wearing_mask_error:
+        return str(not_wearing_mask_error) + f"{masks_to_buy} masks"
```
EdAlekseiev, line 28:
```suggestion
    if not_wearing_mask_error:
        return str(not_wearing_mask_error) + f"{masks_to_buy} masks"
```

EdAlekseiev (changes requested):

VladyslavBazhyn:
I can remove not_wearing_mask_error on the top of my function because it need for the final check. If I remove it with remowing final check program will lose one test from 17.

EdAlekseiev (approved):
Good job

PR #687, opened 2023-11-27
ADone

B1GL3S:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+            if datetime.date.today() > visitor["vaccine"]["expiration_date"]:
+                raise OutdatedVaccineError("Out dated vaccine")
+
+            if not visitor["wearing_a_mask"]:
+                raise NotWearingMaskError("Wear mask please dummy.")
+
+        except KeyError:
+            raise NotVaccinatedError("Visitor should be vaccinated.")
+
+        else:
```
vasylhnatiuk, line 24: else is useless here

app/main.py
```diff
@@ -1 +1,22 @@
...
+        except NotWearingMaskError:
+            masks += 1
+
+        except VaccineError:
+            return "All friends should be vaccinated"
+
+    if masks > 0:
+        return f"Friends should buy {masks} masks"
+
+    else:
```
vasylhnatiuk, line 21: you can avoid else

vasylhnatiuk (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        try:
+            if datetime.date.today() > visitor["vaccine"]["expiration_date"]:
+                raise OutdatedVaccineError("Out dated vaccine")
+
+            if not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 18:
```suggestion
            if not visitor.get("wearing_a_mask"):

```
Use get because wearing_a_mask may not exist

vasylhnatiuk (approved):

PR #693, opened 2023-11-30
Solution

Pavlo-here:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError
+
+        if not visitor["wearing_a_mask"]:
```
ArsenPidhoretskyi, line 21:
```suggestion
        if not visitor.get("wearing_a_mask"):
```
wearing_a_mask may not exist into visitor

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #694, opened 2023-12-01
Solution

eLQeR:


app/main.py
```diff
@@ -1 +1,22 @@
...
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    has_mask = True
+    without_mask = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotVaccinatedError:
+            return "All friends should be vaccinated"
+        except OutdatedVaccineError:
+            return "All friends should be vaccinated"
```
EdAlekseiev, line 16: You can unite these to one `exception` block, because `NotVaccinatedError` and `OutdatedVaccineError` have the same parent class
```suggestion
        except VaccineError:
            return "All friends should be vaccinated"
```

app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError
+        elif not visitor["wearing_a_mask"]:
```
EdAlekseiev, line 16: Use method `get` of dict to avoid `KeyError`

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,20 @@
...
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    has_mask = True
+    without_mask = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            has_mask = False
```
ArsenPidhoretskyi, line 16: This variable is redundant. has_mask is equal to without_mask > 0

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):
Good job

PR #698, opened 2023-12-05
Solution

Ded31:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError")
+        if datetime.date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        if visitor["wearing_a_mask"] is False:
```
EdAlekseiev, line 16: Use method `get` of dict to avoid `KeyError`

EdAlekseiev (changes requested):

PR #708, opened 2023-12-20
Solution

gertsovenko:


app/cafe.py
```diff
@@ -0,0 +1,31 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+
+        if "vaccine" not in visitor:
+            try:
+                raise NotVaccinatedError
+            except NotVaccinatedError:
```
pavlejviki, line 16: there is no need in the try-except block. Raise an error with an informative message

app/cafe.py
```diff
@@ -0,0 +1,31 @@
...
+                raise
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            try:
+                raise OutdatedVaccineError
+            except OutdatedVaccineError:
+                raise
+
+        if not visitor["wearing_a_mask"]:
```
pavlejviki, line 25: use get() method when accessing the value of a key in the dictionary

pavlejviki (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,32 @@
+import datetime
+from app.errors import (
+    NotVaccinatedError,
+    NotWearingMaskError,
+    OutdatedVaccineError
+)
```
Dimosphen1, line 6: Add a blank line to separate different groups of imports according to PEP-8 in the future.

app/cafe.py
```diff
@@ -0,0 +1,32 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(
+                f"Vaccine is required to visit {self.name}"
+            )
+
+        if (visitor.get("vaccine").get("expiration_date")
```
Dimosphen1, line 20: In case you aim to make a more error-tolerant approach using `get`, add a default `dict` that is returned in case `vaccine` key is absent. It will allow the second `.get()` to not fail. 
```suggestion
        if (visitor.get("vaccine", {}).get("expiration_date")

```

Dimosphen1 (approved):
LGTM, well done! Several comments are left for consideration.

PR #709, opened 2023-12-22
Solution

KwintDimon:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
+import datetime
+
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
```
pavlejviki, line 5: opening and closing parenthesis should be on separate lines

pavlejviki (approved):
Good job, just add a minor fix

PR #712, opened 2023-12-28
Solution

ibulakevych:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
EdAlekseiev, line 14: Use method `get` to avoid `KeyError`

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        if visitor["wearing_a_mask"] is False:
```
EdAlekseiev, line 16: Use method `get` to avoid `KeyError`

EdAlekseiev (changes requested):

app/cafe.py
EdAlekseiev: Still not fixed
```suggestion
        if visitor["vaccine"].get("expiration_date") < datetime.date.today():
```

EdAlekseiev (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #716, opened 2024-01-02
Completed the task!

raychw:
Implemented a Cafe class with a following method, necessary exceptions, and a required function successfully.

### DIFF
--- a/app/cafe.py
+++ b/app/cafe.py
@@ -1,0 +1,20 @@
+import datetime
+
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if visitor.get("vaccine") is None:
+            raise NotVaccinatedError
+        if datetime.date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError
+        if visitor["wearing_a_mask"] is False:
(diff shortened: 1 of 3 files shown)


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+
+from app.errors import NotVaccinatedError
+from app.errors import NotWearingMaskError
+from app.errors import OutdatedVaccineError
```
pavlejviki, line 5:
```
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```

pavlejviki: sorry, maybe I wasn't clear. That is what I meant

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if visitor.get("vaccine") is None:
+            raise NotVaccinatedError
+        if datetime.date.today() > visitor["vaccine"].get("expiration_date"):
+            raise OutdatedVaccineError
+        if visitor["wearing_a_mask"] is False:
```
pavlejviki, line 17:
```suggestion
        if not visitor.get("wearing_a_mask"):
```

pavlejviki (changes requested):

app/cafe.py
raychw: oh, okay, i'll fix that

pavlejviki (approved):
Well done!

PR #718, opened 2024-01-02
Solution

dissom:


app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
```suggestion
def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
```

vasylhnatiuk (changes requested):

app/errors.py
```diff
@@ -0,0 +1,18 @@
+class VaccineError(Exception):
+    def __init__(self, *args: object) -> None:
```
vasylhnatiuk, line 2: this `init` is useless. you can write only pass

vasylhnatiuk: do it as well

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #724, opened 2024-01-05
Solution

aLEKS-e3:


app/errors.py
```diff
@@ -0,0 +1,16 @@
+class VaccineError(Exception):
```
NEkropo1, line 1:
```suggestion
class BaseVaccineError(Exception):
```
```suggestion
class AbstractVaccineError(Exception):
```

NEkropo1 (changes requested):

app/errors.py
aLEKS-e3: but what's the difference?

aLEKS-e3:
the tests are written to handle VaccineError... or should I rewrite them?)

msymonovych (approved):
Only thing, maybe you should remove all methods from errors.py.
I belive you should print descriptive messages when you raise the exception (in cafe.py)

app/errors.py
NEkropo1: understanding of abstraction is better with base error, or abstract, I guess we'll refactor this later, since it failing the tests )

app/main.py
```diff
@@ -1 +1,18 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_amount += 1
+
+    if masks_amount:
+        return f"Friends should buy {masks_amount} masks"
+    return f"Friends can go to {cafe.name}"
```
NEkropo1, line 18: I guess Serhii was talking about descriptive messages for classes, which you didn't implement.

NEkropo1 (changes requested):

PR #727, opened 2024-01-07
Solution

dariiashostak:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
+import datetime
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
```suggestion
from app.errors import (
    OutdatedVaccineError,
    NotVaccinatedError,
    NotWearingMaskError
)

```

app/errors.py
```diff
@@ -0,0 +1,14 @@
+class VaccineError(Exception):
+    pass
```
vsmutok, line 2: it is better to write docstring  instead of pass

vsmutok (changes requested):

PR #734, opened 2024-01-09
Solution

msymonovych:


maxbidun:
because not counted in Mate

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor should be vaccinated")
+        if datetime.date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("Vaccine is outdated!")
+        if not visitor["wearing_a_mask"]:
```
aLEKS-e3, line 18: I'd write `elif`, but hope that won't be a big problem)

aLEKS-e3 (approved):

app/cafe.py
NEkropo1: It won't be problem here, since we will rise an Error, but in future it could be some blocks which we are going to change, so better to adhere overall concese logic.

app/errors.py
```diff
@@ -0,0 +1,14 @@
+class VaccineError(Exception):
+    pass
+
+
+class NotVaccinatedError(VaccineError):
+    pass
+
+
+class OutdatedVaccineError(VaccineError):
+    pass
```
NEkropo1, line 10: You should use descriptive messages for these Errors

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (
+    VaccineError,
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    mask_to_buy = 0
```
NEkropo1, line 9:
```suggestion
    missing_masks_count = 0
```
```suggestion
    masks_to_buy
```

NEkropo1 (changes requested):

PR #737, opened 2024-01-11
Solution

kabust:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Visitor has an expired vaccine")
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("Visitor is not wearing a mask")
+        return f"Welcome to {self.name}"
```
msymonovych, line 19: It's not a problem here but useing 'elif' is best practice

app/errors.py
```diff
@@ -0,0 +1,14 @@
...
+class NotVaccinatedError(VaccineError):
+    pass
+
+
+class OutdatedVaccineError(VaccineError):
+    pass
+
+
+class NotWearingMaskError(Exception):
+    pass
```
msymonovych, line 14: You should add descriptive messages for these errors.
I guess it will make it more reusable

msymonovych (approved):

PR #738, opened 2024-01-12
Develop

angelzagniy:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("Should be vaccinated!")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("Your vaccine is expired!")
+        if visitor["wearing_a_mask"] is not True:
```
pavlejviki, line 19: use the get() method when retrieving a key to prevent a KeyError in situations where the key may not be present

pavlejviki (changes requested):

pavlejviki (approved):
Good job, just added a minor suggestion

PR #740, opened 2024-01-13
solution 0.1

PyCore23:


app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from .cafe import Cafe
+from .errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list[dict], cafe: "Cafe") -> None:
```
```suggestion
def go_to_cafe(friends: list[dict], cafe: Cafe) -> None:
```

vasylhnatiuk (changes requested):

app/errors.py
```diff
@@ -0,0 +1,14 @@
+class VaccineError(BaseException):
+    ...
```
EdAlekseiev, line 2: Use `pass` instead of `...`

app/cafe.py
```diff
@@ -0,0 +1,16 @@
+import datetime
+from .errors import (NotVaccinatedError, OutdatedVaccineError, NotWearingMaskError)
```
EdAlekseiev, line 2: use absolute path instead of relative

EdAlekseiev (changes requested):

vsmutok (approved):
Good Job!

PR #751, opened 2024-01-18
Solution

mariia-sergeevna:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | Exception:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor should be vaccinated.")
+        if visitor["vaccine"]["expiration_date"] < date.today():
```
EdAlekseiev, line 17: Use method `get` to avoid `KeyError`
```suggestion
        if visitor["vaccine"].get("expiration_date") < date.today():
```

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+            raise OutdatedVaccineError("The vaccine must not be expired.")
+        if visitor["wearing_a_mask"] is False:
```
EdAlekseiev, line 19:
```suggestion
        if visitor.get("wearing_a_mask"], False):
```

EdAlekseiev (changes requested):

app/cafe.py
mariia-sergeevna: fixed

EdAlekseiev (approved):
Good job

PR #752, opened 2024-01-19
Solution

Meksim-fesh:


app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> (
```
vasylhnatiuk, line 13: this retuns only str. raise is not a return

app/errors.py
```diff
@@ -0,0 +1,18 @@
+class VaccineError(Exception):
+    def __init__(self, *args: object) -> None:
```
vasylhnatiuk, line 2: you can avoid init here and in all of Exception clasess. Just use pass

vasylhnatiuk (changes requested):

vsmutok (approved):
Well Done!

PR #754, opened 2024-01-19
Solution

Disash:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor must be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
EdAlekseiev, line 17: Use method `get` to avoid `KeyError`
```suggestion
        if visitor["vaccine"].get("expiration_date")< datetime.date.today():
```

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+            raise OutdatedVaccineError("Vaccine is expired")
+        if visitor["wearing_a_mask"] is False:
```
EdAlekseiev, line 19:
```suggestion
        if visitor.get("wearing_a_mask"):
```

EdAlekseiev (changes requested):

PR #759, opened 2024-01-22
'Solution'

NataliiaKorzh:


app/main.py
```diff
@@ -1 +1,19 @@
...
+                        NotWearingMaskError)
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
JohnnyWalker010, line 12: You may use VaccineError exception here. We created common class to use it instead of using each of it's children)

JohnnyWalker010 (approved):

app/cafe.py
```diff
@@ -1,7 +1,7 @@
+from datetime import date
 from app.errors import (NotVaccinatedError,
```
NEkropo1, line 2: The section of PEP 8 concerning imports recommends grouping them in the following order:

    Standard library imports.
    Related third-party imports.

app/cafe.py
```diff
@@ -10,10 +10,9 @@ def __init__(self, name: str) -> None:
 
     def visit_cafe(self, visitor: dict) -> str:
         if "vaccine" not in visitor:
-            raise NotVaccinatedError("NotVaccinatedError")
+            raise NotVaccinatedError("This person doesn't have vaccine")
```
NEkropo1, line 13: which person? If you mention some object, link it to error log.

NEkropo1 (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
...
+                        NotWearingMaskError)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
```
MaxOperchuk, line 12: You need to except VaccineError as error and return it like str

MaxOperchuk (changes requested):

PR #765, opened 2024-01-23
Add Cafe class and go_to_cafe function

achernyshovv:


kstorozhenko (approved):
Looks good!

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from typing import Dict
+from app.errors import (
+    NotVaccinatedError, OutdatedVaccineError, NotWearingMaskError)
+import datetime
```
NEkropo1, line 4: fix imports

app/errors.py
```diff
@@ -0,0 +1,22 @@
...
+class OutdatedVaccineError(VaccineError):
+    def __init__(self, message: str = "Visitor's vaccine is outdated") -> None:
+        self.message = message
+        super().__init__(self.message)
+
+
+class NotWearingMaskError(Exception):
+    def __init__(self, message: str = "Visitor is not wearing a mask") -> None:
+        self.message = message
+        super().__init__(self.message)
```
NEkropo1, line 22: why do you need to save this message in self, if you anyway translate it to super?

NEkropo1 (changes requested):

PR #767, opened 2024-01-23
Solution

kstorozhenko:


VladKyianytsia (approved):
Correct!

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+from app.errors import (NotVaccinatedError,
```
NEkropo1, line 2: The section of PEP 8 concerning imports recommends grouping them in the following order:

    Standard library imports.
    Related third-party imports.

NEkropo1 (changes requested):

PR #771, opened 2024-01-23
Solution51

artemgrishko:


app/main.py
```diff
@@ -1 +1,24 @@
...
+from app.errors import (
+    NotWearingMaskError,
+    VaccineError
+)
+
+
+def go_to_cafe(
+        friends: list,
+        cafe: Cafe
+) -> str:
```
tskopenko, line 12: u can use more simplyfiy interface
```python
def go_to_cafe(friends: list,
                         cafe: Cafe) -> str:
```

tskopenko (approved):

PR #776, opened 2024-01-23
Solution

isikach:


app/errors.py
```diff
@@ -0,0 +1,16 @@
+class VaccineError(Exception):
+    def __str__(self) -> str:
+        return "All friends should be vaccinated"
```
Bulleter06, line 3: I think, that you don`t need this.

app/errors.py
```diff
@@ -0,0 +1,16 @@
...
+    pass
+
+
+class OutdatedVaccineError(VaccineError):
+    pass
+
+
+class NotWearingMaskError(Exception):
+    def __str__(self) -> str:
+        return "Friends should buy masks"
```
Bulleter06, line 16: and there

Bulleter06 (changes requested):

app/errors.py
isikach: I return it in visit_cafe function

app/errors.py
isikach: about this - agree

isikach: If I delete it - I will have "AssertionError: NotWearingMaskError should be raised with a message" 
For sure, I can add message to the visit_cafe function, but for me str method looks better

PR #780, opened 2024-01-25
Solution

DenZatolokin:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not ("vaccine" in visitor):
+            raise NotVaccinatedError("Visitor is not vaccinated")
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
EdAlekseiev, line 18: Use method `get` to avoid `KeyError`

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+            raise OutdatedVaccineError("Vaccine has expired")
+
+        if not visitor["wearing_a_mask"]:
```
EdAlekseiev, line 21: The same comment as above
```suggestion
        if not visitor.get("wearing_a_mask"):
```

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #794, opened 2024-02-02
Solution

Artkom98:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("Someone is not vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("The vaccine is not suitable")
+        if not visitor["wearing_a_mask"]:
```
EdAlekseiev, line 16: The same comment as above
```suggestion
        if not visitor.get"wearing_a_mask"):

```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #797, opened 2024-02-04
Solution

voidmain69:


app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
```suggestion
def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
```

app/cafe.py
```diff
@@ -0,0 +1,36 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None | str:
+        try:
+            self.check_vaccination(visitor)
+            self.check_vaccine_expiration(visitor)
+            self.check_wearing_a_mask(visitor)
+        except (NotVaccinatedError, OutdatedVaccineError, NotWearingMaskError):
```
vasylhnatiuk, line 17: you can use only parent class VaccineError instead of 2 child classes

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #798, opened 2024-02-05
solution

MilArtem78:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
+import datetime
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError)
```
Dimosphen1, line 5: Place a parenthesis on new line to match PEP-8

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (VaccineError, NotWearingMaskError)
```
Dimosphen1, line 2: In case you place imports on the same line, there is no need to place them inside of `()`.

Dimosphen1 (changes requested):
Several code styling changes were requested.

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+            name: str
+    ) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError()
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError()
+        if not visitor["wearing_a_mask"]:
```
pavlejviki, line 21: when accessing a key of a dictionary it  is better to use get() method to avoid KeyError if the key does not exist

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #803, opened 2024-02-06
task completed

voronvasia:


app/main.py
```diff
@@ -1 +1,23 @@
...
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_needed = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotVaccinatedError as e:
+            return e.__str__()
```
Dimosphen1, line 15: Do not call dunder methods when it is possible to avoid this. Call `str()` on errors or define the string that should be returned out of the scope of the error implementation.

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
Dimosphen1, line 18: You should call chaining keys in dicts only in cases when you are sure in its structure. In production cases, it is safer to use `get("smth1", {}).get("smth2")` structure.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
vasylhnatiuk, line 9:
```suggestion
def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    masks_needed = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotVaccinatedError as e:
+            return str(e)
+        except OutdatedVaccineError as e:
+            return str(e)
```
Dimosphen1, line 17: There is no need to create two conditions for exceptions, you could catch only `VaccineError` and since both `NotVaccinatedError` and `OutdatedVaccineError` are inherited from it they will be cached and you'll avoid a duplication.

Dimosphen1 (changes requested):

PR #804, opened 2024-02-07
Solution

MykhailoIvchenko:


app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+
+        expiration_date = visitor["vaccine"]["expiration_date"]
```
EdAlekseiev, line 14: Use method get to avoid KeyError
```suggestion
        expiration_date = visitor["vaccine"].get("expiration_date")
```

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+        today = date.today()
+
+        if expiration_date < today:
+            raise OutdatedVaccineError
+
+        if not visitor["wearing_a_mask"]:
```
EdAlekseiev, line 20: The same comment as above

EdAlekseiev (approved):
Good job

app/cafe.py
MykhailoIvchenko: Ok. Thanks. You've given me a very useful advice. But I wouldn't say I like dictionaries in Python at all. JS objects are much better and convenient in use)

PR #811, opened 2024-02-12
Solution

Toltecos:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+from datetime import date
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
```
Dimosphen1, line 6: Add a blank line between different groups of imports according to PEP-8

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if visitor.get("vaccine") is None:
+            raise NotVaccinatedError("The visitor is not vaccinated!")
+        elif visitor.get("vaccine").get("expiration_date") < date.today():
```
Dimosphen1, line 16: Modify this line to prevent calling `get` from `None`
```suggestion
        elif visitor.get("vaccine", {}).get("expiration_date") < date.today():
```

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+            raise OutdatedVaccineError("The vaccination period has expired!")
+        elif visitor.get("wearing_a_mask") is False:
+            raise NotWearingMaskError("The visitor should wear a mask!")
+        else:
```
Dimosphen1, line 20: There is no need for `else`, you could omit it.

Dimosphen1 (changes requested):

PR #829, opened 2024-02-25
Solution

RomanNest:


EdAlekseiev (changes requested):
Fix flake8, please

EdAlekseiev (changes requested):
Fix flake8, please. If you stuck with this problem, you can write to chat or visit QnA

PR #832, opened 2024-03-03
Solution

mish-sk:


app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+
+        elif visitor["vaccine"]["expiration_date"] < date.today():
```
EdAlekseiev, line 18: Use method `get` to avoid `KeyError`.
```suggestion
        elif visitor["vaccine"].get("expiration_date") < date.today():
```

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+            raise OutdatedVaccineError("All friends should have valid vaccine")
+
+        elif not visitor["wearing_a_mask"]:
```
EdAlekseiev, line 21: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #839, opened 2024-03-05
'Solution'

Meyertal:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Cafe:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < date.today():
```
EdAlekseiev, line 16: Use method `get` to avoid KeyError

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+            raise OutdatedVaccineError("All friends should be vaccinated")
+        if visitor["wearing_a_mask"] is False:
```

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+            raise NotWearingMaskError("Friends should buy masks")
+        else:
+            return f"Welcome to {self.name}"
```
EdAlekseiev, line 21:
```suggestion
        return f"Welcome to {self.name}"
```

app/main.py
```diff
@@ -1 +1,19 @@
...
+        try:
+            Cafe.visit_cafe(cafe, person)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            return ("All friends should be vaccinated")
+        except NotWearingMaskError:
+            count += 1
+    if count > 0:
+        return (f"Friends should buy {count} masks")
+    else:
+        return (f"Friends can go to {cafe.name}")
```
EdAlekseiev, line 19:
```suggestion
    if count > 0:
        return (f"Friends should buy {count} masks")
    return (f"Friends can go to {cafe.name}")
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #845, opened 2024-03-06
Solution

mvsydoruk:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+from app.errors import (
```
Dimosphen1, line 2: Add a blank line between different import groups to follow PEP-8

app/main.py
```diff
@@ -1 +1,28 @@
...
+from app.cafe import Cafe
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(cafe: Cafe, friends: list) -> str:
+    can_return = True
```
Dimosphen1, line 10: There is no need for this flag, remove it

app/main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(cafe: Cafe, friends: list) -> str:
```
Dimosphen1, line 9: Indicate that you expect to obtain a list of dict for `friends` in the type annotation

app/main.py
```diff
@@ -1 +1,28 @@
...
+        ) as e:
+            can_return = False
+            if isinstance(e, (NotVaccinatedError, OutdatedVaccineError)):
+                return "All friends should be vaccinated"
+            elif isinstance(e, NotWearingMaskError):
+                masks_to_buy += 1
+    if masks_to_buy > 0:
+        return f"Friends should buy {masks_to_buy} masks"
+    if can_return:
+        return f"Friends can go to {cafe.name}"
```
Dimosphen1, line 28: Add a separate error handling for `NotWearingMaskError` and `VaccineError` (it would be sufficient since both `NotVaccinatedError` and `OutdatedVaccineError`inherit from it) to simplify the logic. You could limit using if condition to only one case at the end of the function

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,23 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except (
+            VaccineError,
+            NotWearingMaskError
+        ) as e:
+            if isinstance(e, VaccineError):
+                return "All friends should be vaccinated"
+            elif isinstance(e, NotWearingMaskError):
+                masks_to_buy += 1
```
Dimosphen1, line 20: Write two except blocks and do not process errors using if conditions, you can manage it in other way
```suggestion
        except VaccineError:
            ....
        except NotWearingMaskError:
            ....
```

Dimosphen1 (changes requested):

vsmutok (approved):
Good Job!

PR #847, opened 2024-03-08
'Solution'

CANR-95:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
EdAlekseiev, line 16: Use method `get` to avoid `KeyError`
```suggestion
        if visitor["vaccine"].get("expiration_date") < datetime.date.today():

```

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        if not visitor["wearing_a_mask"]:
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #852, opened 2024-03-11
Solution

Viesich:


app/cafe.py
```diff
@@ -0,0 +1,23 @@
+import datetime
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
```
Dimosphen1, line 6: Add a blank line between different import groups to follow PEP-8

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    vaccinated = True
```
Dimosphen1, line 6: There is no need for this flag, in case you face a `VaccineError`, you could add a corresponding `return` statement at the same stage, there is no need for further code execution in this case

app/main.py
```diff
@@ -1 +1,22 @@
...
+        except VaccineError:
+            vaccinated = False
+        except NotWearingMaskError:
+            mask_not_wear_count += 1
+
+    if not vaccinated:
+        return "All friends should be vaccinated"
+    elif mask_not_wear_count > 0:
+        return f"Friends should buy {mask_not_wear_count} masks"
+    else:
```
Dimosphen1, line 21: There is no need for `else` since the code will not execute after the `return` statement anyway. Please, remove the redundant `else`

Dimosphen1 (changes requested):

Viesich:
Thank You!

app/main.py
```diff
@@ -1 +1,17 @@
...
+    mask_not_wear_count = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            mask_not_wear_count += 1
+    if mask_not_wear_count > 0:
```
Dimosphen1, line 15: You could make this condition as follows
```suggestion
    if mask_not_wear_count:
```

Dimosphen1 (approved):

PR #854, opened 2024-03-11
Made

antonver:


app/main.py
```diff
@@ -1 +1,26 @@
...
+    try:
+        [cafe.visit_cafe(friend) for friend in friends]
+    except (NotVaccinatedError, OutdatedVaccineError):
+        pass
+    except NotWearingMaskError:
+        masks_to_buy = \
+            sum(1 for friend in friends if not friend["wearing_a_mask"])
+        return f"Friends should buy {masks_to_buy} masks"
+    else:
+        return f"Friends can go to {cafe.name}"
```
Dimosphen1, line 26: You could simplify the logic by introducing `masks_to_buy` flag or another one at the beginning of the function that will be incremented whenever the `NotWearingMaskError` happens. After that, in case the value of the `mask_error_count` does not equal to 0, you can return `f"Friends should buy {masks_to_buy} masks"` or `f"Friends can go to {cafe.name}"` otherwise without additional try-except blocks

app/main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
```
Dimosphen1, line 5: Order your imports in an alphabetical order

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #859, opened 2024-03-13
Solution

okolodka:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+import datetime
```
Dimosphen1, line 7: Order your imports correctly according to PEP-8

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

app/main.py
```diff
@@ -1 +1,17 @@
...
+    no_mask = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            no_mask += 1
+
+    if no_mask > 0:
```
Dimosphen1, line 15: Do not use comparison with zero in the future, you could leave
```suggestion
    if no_mask:
```

PR #865, opened 2024-03-15
'Solution'

ArtomUmantsev:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+        elif visitor["vaccine"].get("expiration_date",
+                                    datetime.date.min) < datetime.date.today():
+            raise OutdatedVaccineError("Vaccine is outdate")
+        elif visitor["wearing_a_mask"] is False:
```
EdAlekseiev, line 17: Use `get` method to avoid KeyError
```suggestion
        elif not visitor.get("wearing_a_mask"):

```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #869, opened 2024-03-17
Dev

Bujhm99:


app/main.py
```diff
@@ -1 +1,21 @@
...
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            problems = True
+            no_mask += 1
+
+    if no_mask:
+        return f"Friends should buy {no_mask} masks"
+
+    if not problems:
+        return f"Friends can go to {cafe.name}"
```
Boyok1337, line 21: you can make it without "problems". use "no_mask" > 1 to return f"Friends should buy {no_mask} masks" and else: return f"Friends can go to {cafe.name}"
remove problems variable

app/main.py
```diff
@@ -1 +1,18 @@
...
+        try:
+            cafe.visit_cafe(human)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            no_mask += 1
+
+    if no_mask:
+        return f"Friends should buy {no_mask} masks"
+    else:
```
YuriiKindrat, line 17: why do u need else here?)

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ!

PR #871, opened 2024-03-17
Solution

rosttv:


app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("Visitor should be vaccinated!")
+        expiration_date = visitor.get("vaccine").get("expiration_date")
```
Dimosphen1, line 17: It is a safer way to provide a default value that has an implemented get method here
```suggestion
        expiration_date = visitor.get("vaccine", {}).get("expiration_date")
```

Dimosphen1 (approved):

PR #875, opened 2024-03-19
'Solution'

Vitalii-Haponiuk:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
pavlopro, line 17: It would also be good to check if expiration_date exists and, if it does, what value it holds, to prevent a comparison error.

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+            raise OutdatedVaccineError("The visitor's vaccine is outdated")
+        elif not visitor["wearing_a_mask"]:
```
pavlopro, line 19: The same applies here. It would be better to use the get method to prevent a KeyError.

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
```
pavlopro, line 2: Please format the import of errors as in `cafe.py`.

app/main.py
```diff
@@ -1 +1,16 @@
...
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
pavlopro, line 5: Please also enhance the friends annotation to list of dictionaries

pavlopro (changes requested):

PR #877, opened 2024-03-19
'Solution'

o-zinevych:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("The visitor is not vaccinated.")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
EdAlekseiev, line 16: Use `get` method to avoid `KeyError`
```suggestion
        if visitor["vaccine"].get("expiration_date") < datetime.date.today():

```

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+            raise OutdatedVaccineError("The vaccine is expired.")
+        if visitor["wearing_a_mask"] is False:
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
```suggestion
def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
```

vasylhnatiuk (changes requested):

PR #879, opened 2024-03-20
'Solution'

liudmylaperchevska:


app/cafe.py
```diff
@@ -0,0 +1,23 @@
+from datetime import date
```
pavlopro, line 1: Add blank line between different import types

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        else:
+            if visitor["vaccine"]["expiration_date"] < date.today():
```
pavlopro, line 17: It would be great to check if expiration_date exists and, if it does, check its value to prevent a comparison error

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        else:
```
pavlopro, line 16: No need to have this else statement.
Just keep checking

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+            if visitor["vaccine"]["expiration_date"] < date.today():
+                raise OutdatedVaccineError
+
+            else:
+                if not visitor["wearing_a_mask"]:
```
pavlopro, line 21: It would be great to check if `wearing_a_mask `exists in visitor to prevent KeyError

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
pavlopro, line 5: Adjust friends type annotation to list if dicts

app/errors.py
```diff
@@ -0,0 +1,17 @@
+class VaccineError(Exception):
+    pass
+
+
+class NotVaccinatedError(VaccineError):
+    def __init__(self) -> None:
+        super().__init__("Visitor is not vaccinated")
```
pavlopro, line 7: There's no need to redefine the __init__ method. Just pass the message directly when calling the exception, 
like this: `raise NotVaccinatedError("Visitor is not vaccinated")`

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #880, opened 2024-03-20
create new exceptions and cafe class, and function

MaxSEngineer:


app/main.py
MaxSEngineer, line 1: delete break

vsmutok (approved):
Looks Good!

PR #884, opened 2024-03-24
Solution

prev1ew:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
+import datetime
+
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError, NotWearingMaskError)
```
Dimosphen1, line 4: Place each import on a new line so there would be no imports on the same line with opening and closing parenthesis.

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
```
Dimosphen1, line 12: Simplify this condition since it is looking for th e presence of the string in the keys by default
```suggestion
        if "vaccine" not in visitor:
```

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+            raise NotVaccinatedError
+
+        expiration_date = visitor["vaccine"].get("expiration_date")
+        if (expiration_date is None
+                or expiration_date < datetime.date.today()):
+            raise OutdatedVaccineError
+
+        visitor_wears_a_mask = visitor.get("wearing_a_mask")
+        if (visitor_wears_a_mask is None) or (not visitor_wears_a_mask):
```
Dimosphen1, line 21: Do not create `visitor_wears_a_mask` variable
```suggestion
        if not visitor.get("wearing_a_mask"):
```

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+
+def go_to_cafe(friends: List[Dict], cafe: Cafe) -> str:
+    not_all_vaccinated = False
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            not_all_vaccinated = True
```
Dimosphen1, line 14: Place the return statement here and do not use `not_all_vaccinated` flag

Dimosphen1 (changes requested):

PR #888, opened 2024-03-25
cafe_1

Mirinness:


cafe.py
```diff
@@ -0,0 +1,30 @@
+from datetime import date
+from typing import Dict, Union
+from errors import NotVaccinatedError, \
+    OutdatedVaccineError, NotWearingMaskError
```
vsmutok, line 4: better to use `()` then `\`

cafe.py
```diff
@@ -0,0 +1,30 @@
...
+        self.name = name
+
+    def visit_cafe(self,
+                   visitor:
+                   Dict
+                   [
+                       str, Union[str, int, Dict[str, Union[date, None]], bool]
+                   ]
+                   ) \
+            -> str:
```
vsmutok, line 18: so bad style of code

vsmutok (changes requested):

cafe.py
Mirinness: I understand

PR #891, opened 2024-03-28
Dev

CharmNiovi:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app import errors
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+
+    for friend in friends:
+
```
oleksandra-shershen, line 9: Remove the unnecessary line

app/cafe.py
```diff
@@ -0,0 +1,25 @@
+from app import errors
```
oleksandra-shershen, line 1: In my opinion, it's better to use:
```
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+from datetime import date
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | None:
```
oleksandra-shershen, line 9: You can do it like this:
`def visit_cafe(self, visitor: dict) -> str | NotVaccinatedError:`

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+            raise errors.NotVaccinatedError(
+                f"{visitor['name']} is not vaccinated!"
+            )
+
+        if date.today() > visitor["vaccine"]["expiration_date"]:
+            raise errors.OutdatedVaccineError(
+                f"Outdated vaccine for {visitor['name']}"
+            )
+
+        if not visitor.get("wearing_a_mask") is True:
```
oleksandra-shershen, line 20: A simpler and easier option:
` if visitor["wearing_a_mask"] is False: `

oleksandra-shershen (changes requested):
Check comments!)

app/cafe.py
CharmNiovi: In this case we can pass "False", then it will pass

oleksandra-shershen (approved):
GJ

app/cafe.py
```diff
@@ -0,0 +1,26 @@
+from datetime import date
+
+from app import errors
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | None:
```
Oleksl888, line 10: The method returns `str` type, raising an error does not imply that functions return type will be `None`

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+            raise errors.NotVaccinatedError(
+                f"{visitor['name']} is not vaccinated!"
+            )
+
+        if date.today() > visitor["vaccine"]["expiration_date"]:
+            raise errors.OutdatedVaccineError(
+                f"Outdated vaccine for {visitor['name']}"
+            )
+
+        if not visitor.get("wearing_a_mask") is True:
```
Oleksl888, line 21: This statement is overcomplicated

app/cafe.py
```diff
@@ -0,0 +1,26 @@
+from datetime import date
+
+from app import errors
```
Oleksl888, line 3: Your code below will look cleaner if you explicitly import errors. This will improve readabllity and will remove necessity to write `error.ErrorName `

Oleksl888 (changes requested):

PR #893, opened 2024-03-28
Solution

y-kondrashova:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
+from datetime import date
+
+from app.errors import NotWearingMaskError
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
```
DmytroHlazyrin, line 5: You can list all necessary classes that are in one module, comma separated.

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError
+from app.errors import NotWearingMaskError
```
DmytroHlazyrin, line 3: It's the same here

app/errors.py
```diff
@@ -0,0 +1,14 @@
+class VaccineError(Exception):
+    pass
+
+
+class NotVaccinatedError(VaccineError):
+    pass
```
DmytroHlazyrin, line 6: You can add message for error using magic method str. But it's not necessary.

DmytroHlazyrin (changes requested):
Good job. Just need to fix the little mistakes)

YuriiKindrat (approved):
GJ!

PR #894, opened 2024-03-28
Solution

DmytroHlazyrin:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        elif datetime.date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError
+        elif "wearing_a_mask" not in visitor or not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError
+        else:
+            return f"Welcome to {self.name}"
```
panicua, line 19: ```else:``` is unneeded

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (VaccineError,
+                        NotWearingMaskError)
```
panicua, line 3: I don't think you need to break line here, the line isn't that long:
```from app.errors import VaccineError, NotWearingMaskError```

File: app/main.py [hunk 3]
```diff
@@ -1 +1,16 @@
...
+    masks_to_buy = 0
+    for visitor in friends:
+        try:
+            cafe.visit_cafe(visitor)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    return f"Friends should buy {masks_to_buy} masks" if masks_to_buy \
+        else f"Friends can go to {cafe.name}"
```
panicua, line 16: Breaking a line with ```\``` is making code less readable.
You can add parenthases ```()``` around your ```return``` so you can brake a line without ```\```, but it's your decision if you think it's not making your code less readable, don't change it.

panicua (changes requested):

YuriiKindrat (approved):
GJ!

PR #896, opened 2024-03-28
'Solution'

March1205:
I don't understand why I keep getting this error Traceback:
..\AppData\Local\Programs\Python\Python312\Lib\importlib\__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests\test_main.py:6: in <module>
    from app import main
app\main.py:2: in <module>
    from cafe import Cafe
E   ModuleNotFoundError: No module named 'cafe'
I did everything right, created python files in app, but it still can't see them. What am I missing?

### DIFF
--- a/app/cafe.py
+++ b/app/cafe.py
@@ -1,0 +1,20 @@
+import datetime
+from errors import NotVaccinatedError, OutdatedVaccineError, NotWearingMaskError
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        vaccine = visitor.get("vaccine")
+        if not vaccine:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+
+        expiration_date = vaccine.get("expiration_date")
+        if expiration_date and expiration_date < datetime.date.today():
+            raise OutdatedVaccineError("Visitor's vaccine is outdated")
+
+        if not visitor.get("wearing_a_mask", True):
+            raise NotWearingMaskError("Visitor is not wearing a mask")
+        return f"Welcome to {self.name}"
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,15 @@
-# write your code here
+import datetime
+from cafe import Cafe
+from errors import NotVaccinatedError, OutdatedVaccineError, NotWearingMaskError
+
+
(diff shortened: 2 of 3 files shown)


EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,25 @@
...
+
+def go_to_cafe(friends: List[dict], cafe: Cafe) -> str:
+    vaccinated_friends = 0
+    masks_to_buy = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+            vaccinated_friends += 1
+        except NotVaccinatedError:
```
vasylhnatiuk, line 15: You should use VaccineError instead of duplicating of code

app/main.py
```diff
@@ -1 +1,25 @@
...
+            return "All friends should be vaccinated"
+        except OutdatedVaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
+        return f"Friends should buy {masks_to_buy} masks"
+    else:
```
vasylhnatiuk, line 24: You can avoid else. Do it

vasylhnatiuk (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,20 @@
+from datetime import date
+from app.errors import (NotWearingMaskError,
+                        NotVaccinatedError, OutdatedVaccineError)
```
vsmutok, line 3:
```suggestion
from app.errors import (
    NotWearingMaskError,
    NotVaccinatedError, 
    OutdatedVaccineError
)
```

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from typing import List
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: List[dict], cafe: Cafe) -> str:
+    vaccinated_friends = 0
```
vsmutok, line 7: Why you need this variable?

vsmutok (changes requested):

app/main.py
March1205: At first I though I needed it to keep track of friends, but now I see that it is indeed useless

vsmutok (approved):
Looks Good!

PR #899, opened 2024-03-28
Solution for lockdown cafe

mikhailyemets:


Gilgumish (approved):
GJ

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None | str:
```
LiudmylaKulinchenko, line 12: There is no case in the function that will make it return None. You can change return annotation to `-> str`

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+                " Cant let you in"
+            )
+        elif (visitor["wearing_a_mask"] is False
+              or visitor["wearing_a_mask"] is None):
+            raise NotWearingMaskError(
+                "You need to wear a mask."
+                " Cant let you in without it"
+            )
+        else:
+            return f"Welcome to {self.name}"
```
LiudmylaKulinchenko, line 27: Else is unnecessary with only return in it

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None | str:
+        if visitor.get("vaccine") is None:
```
LiudmylaKulinchenko, line 13: `if not visitor.get("vaccine")` will also work here. Same for other condition

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+            raise NotVaccinatedError("You've got no vaccine. Cant let you in.")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError(
+                "Your vaccine has been expired."
+                " Cant let you in"
+            )
+        elif (visitor["wearing_a_mask"] is False
```
LiudmylaKulinchenko, line 20: Use `not` instead of `is False`

app/main.py
```diff
@@ -1 +1,20 @@
...
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    return (
+        f"Friends should buy {masks_to_buy} masks" if masks_to_buy
+        else f"Friends can go to {cafe.name}"
+    )
```
LiudmylaKulinchenko, line 20: Avoiding usage of ternary operator will make it more readable. It will also take one line less

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 💪

PR #900, opened 2024-03-29
Solution

Gilgumish:


SashaKisliy (approved):
GJ!

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+import datetime
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | None:
```
LiudmylaKulinchenko, line 11: There is no case that your function will return None. You should fix your annotation

app/main.py
```diff
@@ -1 +1,18 @@
...
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 1:
```
LiudmylaKulinchenko, line 15: What if masks_to_buy equal 1? Then friends also should go buy masks
You can use just `if masks_to_buy` here, since it will return True only if masks_to_buy > 0

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Visitor's vaccine is outdated")
+
+        if visitor["wearing_a_mask"] is False:
```
LiudmylaKulinchenko, line 18: Use `not` instead of `is False`

LiudmylaKulinchenko (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
+from app.errors import (NotVaccinatedError,
```
Oleksl888, line 1: Check the order of imports: what should come first? In which order?

Oleksl888 (changes requested):

Oleksl888 (approved):
Approved, next time sort imports in alphabetical orders to make it perfect

PR #901, opened 2024-03-29
Solution

SashaKisliy:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+
+import datetime
```
vitalinamal, line 5: It is best practice to write built-in modules first

LiudmylaKulinchenko: Fix imports order

LiudmylaKulinchenko (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+
+from app.errors import (NotVaccinatedError,
```
Oleksl888, line 3: It is also a best practice to sort imports in alphabetical order. Check below as well

Oleksl888 (changes requested):

Oleksl888 (approved):

PR #904, opened 2024-03-29
'Solution'

kssshevchenko:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor must be vaccinated")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Visitor's vaccine is expired")
+        if not visitor.get("wearing_a_mask", False):
+            raise NotWearingMaskError("Visitor must be wearing a mask")
```
panicua, line 17: I belive this condition is overcomplicated, did you try:
```if not visitor.get("wearing_a_mask"):``` ?

File: app/main.py [hunk 2]
```diff
@@ -1 +1,19 @@
...
+                        NotWearingMaskError)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            return "All friends should be vaccinated"
```
panicua, line 13: We were creating class ```VaccineError``` so we don't need to catch those errors, we just need to catch their parent class error.

app/main.py
```diff
@@ -1 +1,19 @@
...
+        except NotWearingMaskError:
+            masks_to_buy += 1
+            continue
```
panicua, line 16: I believe this ```continue``` keyword does nothing to code, or I'm wrong?

panicua (changes requested):

YuriiKindrat (approved):
GJ

PR #906, opened 2024-03-29
'Solution'

yossosh:


Chub-andrew (approved):

app/main.py
```diff
@@ -1 +1,28 @@
...
+            except NotVaccinatedError:
+                return "All friends should be vaccinated"
+            except OutdatedVaccineError:
+                return "All friends should be vaccinated"
+            except NotWearingMaskError:
+                masks_to_buy += 1
+
+        if vaccinated_friends == len(friends):
+            return f"Friends can go to {cafe.name}"
+        else:
```
YuriiKindrat, line 25: why do u need else here ?
also u could make condition like this 
`if masks_to_buy:`

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Good job ☀

PR #907, opened 2024-03-29
Implement Lockdown cafe task

bwaah1:


app/main.py
```diff
@@ -1 +1,21 @@
...
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    not_vaccinated = False
+    masks_count = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            not_vaccinated = True
```
YuriiKindrat, line 13: could u return "All friends should be vaccinated" here, so there is won't be need to continue interate through list

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #909, opened 2024-03-30
Solution

Chub-andrew:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Vaccine is not")
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError(visitor["vaccine"]["expiration_date"])
+        if visitor["wearing_a_mask"] is False:
+            raise NotWearingMaskError("Wearing mask")
+        if not visitor.get("wearing_a_mask"):
+            raise NotWearingMaskError("Wearing mask")
```
alonasorochynska, line 19: You can reduce code in this method. Use elif to check if vaccine still valid, ese elif to check if mask is True or False (here can be construction if, elif, elif)

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+                        NotVaccinatedError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Vaccine is not")
```
alonasorochynska, line 13: Vaccine is not what? :)

app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+from app.errors import (OutdatedVaccineError,
```
alonasorochynska, line 2: if I remember well between imports of different types of libs we have to insert 2 empty lines

alonasorochynska (changes requested):
You can improve your code!

YuriiKindrat (approved):
GJ

PR #912, opened 2024-03-31
Solution#1

VitalyBashkiser:


yossosh (approved):
good job!

app/cafe.py
```diff
@@ -0,0 +1,21 @@
+from app.errors import NotVaccinatedError, OutdatedVaccineError,\
+    NotWearingMaskError
```
YuriiKindrat, line 2: no need to use linebreaker here, use brackets instead

YuriiKindrat (changes requested):

YuriiKindrat (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,24 @@
+from app.errors import (
+    NotVaccinatedError,
+    NotWearingMaskError,
+    OutdatedVaccineError,
+)
+from datetime import date
```
Oleksl888, line 6: order of imports here?

Oleksl888 (changes requested):

Oleksl888 (approved):
well done!

PR #914, opened 2024-04-01
Solution

siaserb:


afaidiuk (approved):
Great!

app/cafe.py
```diff
@@ -0,0 +1,32 @@
...
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | None:
+        exception_counter = 0
```
YuriiKindrat, line 13: why do u need this variable ? it seems to me, the work of method will be interupted in case of error

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Well done 🌸

PR #915, opened 2024-04-01
'Solution'

afaidiuk:


app/cafe.py
```diff
@@ -0,0 +1,27 @@
+from datetime import date
+
+from app.errors import (
```
Oleksl888, line 3: It is a good practice to sort imports in alphabetical order

app/main.py
```diff
@@ -1 +1,28 @@
...
+    masks_to_buy = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+            vaccinated_count += 1
+        except NotVaccinatedError:
+            return "All friends should be vaccinated"
+        except OutdatedVaccineError:
+            return "All friends should be vaccinated"
```
Oleksl888, line 21: These two errors inherit from the same class. So how can you simplify these two statements?

PR #918, opened 2024-04-02
'Solution'

dkotkod:


app/main.py
```diff
@@ -1 +1,16 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy > 0:
```
Oleksl888, line 14: `if masks_to_buy` is enough, don't compare a variable to 0 if there is no need

app/cafe.py
```diff
@@ -0,0 +1,22 @@
+import datetime
+from app.errors import (
```
Oleksl888, line 2: should be a blank line between different types of imports. Also it is a good practice to imports alphabetically

Oleksl888 (changes requested):

Oleksl888 (approved):
good job!

PR #919, opened 2024-04-03
PC-lockdown-cafe

DOLC3-GABBANA:


app/cafe.py
```diff
@@ -0,0 +1,27 @@
+import datetime
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
```
Dimosphen1, line 6: Add a blank line between different groups of imports to follow PEP-8 in all files where applicable

app/main.py
```diff
@@ -1 +1,44 @@
...
+                    )
+
+            else:
+                raise OutdatedVaccineError(
+                    f"Incompletevaccine "
+                    f"information for {friend['name']}"
+                )
+
+            if not friend.get("wearing_a_mask", False):
+                masked_friends += 1
```
Dimosphen1, line 36: Do not duplicate code, use defined `visit_cafe` method here

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+        except (NotVaccinatedError, OutdatedVaccineError) as e:
```
Dimosphen1, line 38: You could import and catch only `VaccineError` since it is a parent of both `NotVaccinatedError` and `OutdatedVaccineError`

app/main.py
```diff
@@ -1 +1,44 @@
...
+            return str(e)
+
+    if masked_friends > 0:
+        return f"Friends should buy {masked_friends} masks"
+    else:
```
Dimosphen1, line 43: There is no need for `else` since the code will not execute after the `return` statement anyway. Please, remove the redundant `else`

app/main.py
```diff
@@ -1 +1,44 @@
...
+                        f"vaccine has expired"
+                    )
+
+            else:
+                raise OutdatedVaccineError(
+                    f"Incompletevaccine "
+                    f"information for {friend['name']}"
+                )
+
+            if not friend.get("wearing_a_mask", False):
```
Dimosphen1, line 35: You could not define `False` here since the default `None` returned by `get` is faulty by boolean type

Dimosphen1 (changes requested):

EdAlekseiev (approved):

PR #924, opened 2024-04-06
solved

serhii-chaban:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+            vaccine = visitor["vaccine"]
+            assert vaccine["expiration_date"] >= datetime.date.today()
+        except KeyError:
+            raise NotVaccinatedError(f"{visitor['name']} has no vaccine!")
+        except AssertionError:
+            raise OutdatedVaccineError(f"{visitor['name']}'s "
+                                       f"vaccine is expired!")
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError(f"{visitor['name']} has no mask!")
+        return f"Welcome to {self.name}"
```
vsmutok, line 25:
```suggestion
    def visit_cafe(self, visitor: dict) -> str:
        if "vaccine" not in visitor:
            raise NotVaccinatedError("All friends should be vaccinated")
        today = datetime.date.today()
        expiration_date = visitor["vaccine"]["expiration_date"]
        if today > expiration_date:
            raise OutdatedVaccineError("All friends should be vaccinated")
        if not visitor["wearing_a_mask"]:
            raise NotWearingMaskError("Friends should buy masks")
        return f"Welcome to {self.name}"
```

vsmutok (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+        today = datetime.date.today()
+        expiration_date = visitor["vaccine"]["expiration_date"]
+        if today > expiration_date:
```
Dimosphen1, line 19: There is no need to create variables since they are used only once in the code. Evaluate the expression as is without variable assignment in the future if possible.

Dimosphen1 (approved):

PR #925, opened 2024-04-06
Solution

jannhol:


app/main.py
```diff
@@ -1 +1,22 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy:
+        return f"Friends should buy {masks_to_buy} masks"
+    else:
```
Dimosphen1, line 21: There is no need for `else` since the code will not execute after the `return` statement anyway. Please, remove the redundant `else`

Dimosphen1 (changes requested):

EdAlekseiev (approved):

PR #929, opened 2024-04-07
Solution

Ruxxell54:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+import datetime
```
Oleksl888, line 1: must be blank lines between different import types.
import also must be sorted alphabetically

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError, NotWearingMaskError)
+
+
+class Cafe:
+    name = ""
```
Oleksl888, line 7: why do you use class attribute here?

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("something 3")
```
Oleksl888, line 14: please make error messages more descriptive

app/main.py
```diff
@@ -1 +1,16 @@
...
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_required = 0
+    for person in friends:
+        try:
+            cafe.visit_cafe(person)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_required = masks_required + 1
```
Oleksl888, line 13: use +=1 in cases like these.

app/main.py
```diff
@@ -1 +1,16 @@
...
+    if masks_required != 0:
```
Oleksl888, line 14: Here you can simply use `if not masks_required` because it will return falsy value

Oleksl888 (changes requested):

Oleksl888 (approved):
Nice!

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Cafe:
+    def __init__(self, cafe_name: str) -> None:
+        self.name = cafe_name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated.")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
```
Oleksl888, line 15: you need to check all 3 conditions all the time thats why it must be `if` not `elif`

PR #950, opened 2024-04-29
"Solution"

poplyasita:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+
+from app.errors import (
+    NotVaccinatedError, OutdatedVaccineError, NotWearingMaskError
```
Dimosphen1, line 4: Place each import on a new line to improve the readability

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: [dict], cafe: Cafe) -> None | str:
```
Dimosphen1, line 5: There are no cases when `None` is returned here, leave only `str` in the return type annotation. Besides that, indicate list explicitly in the type annotation
```suggestion
def go_to_cafe(friends: list[dict], cafe: Cafe) -> None | str:
```

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None | str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("NotVaccinatedError")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("OutdatedVaccineError")
+        elif not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("NotWearingMaskError")
+        else:
```
Dimosphen1, line 19: There is no need for `else` since the code will not execute after the raised exceptions anyway. Please, remove the redundant `else`

Dimosphen1 (changes requested):

PR #979, opened 2024-05-29
Solution

b4oody:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from datetime import date
```
pavlopro, line 1: Add blank line between different import groups

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Vaccine is not")
+        elif visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("Vaccine is outdated")
+        elif not visitor["wearing_a_mask"]:
```
pavlopro, line 16: It would be better to check it with the get method, as 'wearing_a_mask' might be absent in the visitor dictionary

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+            raise NotWearingMaskError("Wearing mask is not")
+        else:
```
pavlopro, line 18: else is redundant here

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.errors import NotWearingMaskError, VaccineError
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    any_issues = False
```
pavlopro, line 6: There is no need for this 'any_issues' variable. Remove it and simply `return f"Friends can go to {cafe.name}"` at the end.

pavlopro (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from datetime import date
+
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
```
vsmutok, line 5:
```suggestion

from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.errors import NotWearingMaskError, VaccineError
+
```
Dimosphen1, line 2: There is no need to add a blank line within imports from the same group

Dimosphen1 (approved):

PR #980, opened 2024-05-29
'Solution'

AlexTarasov57:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (
+    VaccineError,
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
vasylhnatiuk, line 8:
```suggestion
def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
```

vasylhnatiuk (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(
+            self,
+            visitor: dict
+    ) -> str | VaccineError | NotWearingMaskError:
```
EdAlekseiev, line 17: Your function doesn't return VaccineError and  NotWearingMaskError. We shouldn't show possible errors in return type hint
```suggestion
    ) -> str:

```

EdAlekseiev (changes requested):

AlexTarasov57:
This all!

PR #982, opened 2024-05-30
Solution

TsvetanKichuk:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
+from datetime import date
+
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
```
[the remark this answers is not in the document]
TsvetanKichuk: Would it be correct if each import was on a separate line?

app/main.py
```diff
@@ -1 +1,26 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotVaccinatedError:
+            return "All friends should be vaccinated"
+        except OutdatedVaccineError:
+            return "All friends should be vaccinated"
```
EdAlekseiev, line 19: You can join these 2 handlers to one, because `NotVaccinatedError` and `OutdatedVaccineError` have the same parent
```suggestion
        except VaccineError:
            return "All friends should be vaccinated"
```

app/main.py
```diff
@@ -1 +1,26 @@
...
+        except NotWearingMaskError:
+            if not friend.get("wearing_a_mask"):
```
EdAlekseiev, line 21: Why do you check it here?

EdAlekseiev (changes requested):

app/main.py
TsvetanKichuk: where I must to check

Polyakiv-Andrey (approved):
Good job!

PR #984, opened 2024-05-31
solution

VShveda:


DevSpaciX (changes requested):
All good , add doc sting under the exceptions

PR #986, opened 2024-06-01
'Solution'

BohdanKuzik:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Vaccine is expired")
+        elif "wearing_a_mask" not in visitor or not visitor["wearing_a_mask"]:
```
DevSpaciX, line 16: You can use `get()` here

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+            raise NotWearingMaskError("Visitor is not wearing a mask")
+        else:
+            return f"Welcome to {self.name}"
```
DevSpaciX, line 19: You can remove redundat else operator

DevSpaciX (changes requested):

Oleg-s-s (approved):
👍

PR #996, opened 2024-06-03
Solution

Andrii1000:


app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+
+        if date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("All friends should be vaccinated")
+
+        if visitor["wearing_a_mask"] is False:
+            raise NotWearingMaskError("All visitors must wear masks")
```
DevSpaciX, line 21: add descriptive error message; an error message is strange. You perform checks for one visitor, so if he is not vaccinated why the message is "all friends". Who are these friends?
I mean use visitor name for better error message : )

DevSpaciX (changes requested):

DevSpaciX (approved):
Good job!

PR #1001, opened 2024-06-03
Solution

SemenetsSergii:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.errors import VaccineError
+from app.errors import NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: str) -> str:
```
DevSpaciX, line 5: Are you sure. that cafe it is `str` ?)

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.errors import VaccineError
+from app.errors import NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: str) -> None:
```
DevSpaciX, line 5: No-no , that you returning `str` - was correct , `cafe` is not `str` type . It is Cafe class type is not it ?

DevSpaciX (changes requested):

DevSpaciX (changes requested):
def go_to_cafe(friends: list, cafe: Сafe) -> str:
Seems like that it should be

Oleg-s-s (approved):
👍

PR #1003, opened 2024-06-03
Solution

b3v3kt0r:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("All friends should be vaccinated")
```
DevSpaciX, line 14: add descriptive error message; an error message is strange. You perform checks for one visitor, so if he is not vaccinated why the message is "all friends". Who are these friends? 
I mean use visitor name for better error message : )

DevSpaciX (changes requested):

b3v3kt0r:
> All tests in GithubAction should pass. For now you have `flake8` errors

in pycharm it works, here - not

DevSpaciX (approved):

PR #1031, opened 2024-06-23
cafe visiting done

Ppllupold:


app/main.py
```diff
@@ -1 +1,17 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            not_wearing_mask_count += 1
+    if not_wearing_mask_count != 0:
+        return f"Friends should buy {not_wearing_mask_count} masks"
+    else:
+        return f"Friends can go to {cafe.name}"
```
vsmutok, line 17:
```suggestion
    if not_wearing_mask_count != 0:
        return f"Friends should buy {not_wearing_mask_count} masks"
    return f"Friends can go to {cafe.name}"
```
`else` redundant in this case

app/cafe.py
```diff
@@ -0,0 +1,25 @@
+from datetime import date
+
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError, NotWearingMaskError)
```
```suggestion
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError, 
    NotWearingMaskError
)
```

vsmutok (changes requested):

PR #1032, opened 2024-06-24
+

AndriZhok:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
Dimosphen1, line 3: Replace this import with the following
```suggestion
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```
There is no need to duplicate `from app.errors import`

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+import datetime
```
Dimosphen1, line 6: Order your imports correctly to follow PEP-8 import rules

app/cafe.py
```diff
@@ -0,0 +1,25 @@
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
+
```
Dimosphen1, line 4: There is no need to add two blank lines between different import groups, one is enough

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, cafe_name: str) -> None:
+        self.cafe_name = cafe_name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+
+        vaccine_info = visitor["vaccine"]
+        expiration_date = vaccine_info["expiration_date"]
+        if expiration_date < datetime.date.today():
```
Dimosphen1, line 19: There is no need to define multiple arguments, perform retrieval in the if condition instead

app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+from app.cafe import Cafe
+
```
Dimosphen1, line 2: There is no need to separate imports from the same import group with blank line

app/main.py
```diff
@@ -1 +1,24 @@
...
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
Dimosphen1, line 5: Change the import according to previous recommendations

app/main.py
```diff
@@ -1 +1,24 @@
...
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotVaccinatedError:
+            return "All friends should be vaccinated"
+        except OutdatedVaccineError:
+            return "All friends should be vaccinated"
```
Dimosphen1, line 16: Think about how you can avoid this code duplication. Pay attention to exception inheritance hierarchy

app/main.py
```diff
@@ -1 +1,24 @@
...
+        except NotWearingMaskError as e:
+            if str(e) == "Visitor's not wearing a mask":
```
Dimosphen1, line 18: There is no need to check the string here, since it is the only option for `NotWearingMaskError `

app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+from app.cafe import Cafe
+
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
Dimosphen1, line 8: Indicate a list of what instances you expect to obtain for `friends` in the type annotation

Dimosphen1 (changes requested):

pavlopro (changes requested):
Please take into consideration all of previous comments

PR #1052, opened 2024-07-15
implemented quarantine cafe checking for vaccine

MasakDirt:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
+import datetime
+
+from app.errors import (NotVaccinatedError, OutdatedVaccineError,
```
oliasymak24, line 3: Will be better to put each class from new line.

oliasymak24 (approved):
Good job!

PR #1055, opened 2024-07-18
Implement solution

anton-bilous:


app/main.py
```diff
@@ -1 +1,16 @@
...
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    missing_masks = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            missing_masks += 1
+    if missing_masks == 0:
```
MNDonut, line 14: use `not` instead of zero comparison

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from .cafe import Cafe
+from .errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    missing_masks = 0
+    for friend in friends:
```
oliasymak24, line 7: I suggest separate the loop from two sides with an empty line for better readability

oliasymak24 (approved):

PR #1058, opened 2024-07-18
solution

Morphin20th:


app/errors.py
```diff
@@ -0,0 +1,16 @@
+class VaccineError(Exception):
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}"
```
vsmutok, line 3: it's not very informative, add a clearer message

app/errors.py
```diff
@@ -0,0 +1,16 @@
...
+    """Person is not vaccinated"""
+
+
+class OutdatedVaccineError(VaccineError):
+    """The vaccine term has expired"""
+
+
+class NotWearingMaskError(Exception):
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}"
```
vsmutok, line 16: it's not very informative, add a clearer message

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
...
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (VaccineError):
```
Dimosphen1, line 13: There is no need to put error inside `()`, you could omit them
```suggestion
        except VaccineError:
```

app/errors.py
```diff
@@ -0,0 +1,16 @@
...
+    """Person is not vaccinated"""
+
+
+class OutdatedVaccineError(VaccineError):
+    """The vaccine term has expired"""
+
+
+class NotWearingMaskError(Exception):
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}: Person is not wearing a mask"
```
Dimosphen1, line 16: There is no need to add a class name in the string method, the exception class would be seen in a traceback in case raised. Please, remove it. Besides that, it would be better to avoid linking string descriptions to a particular class or name to which is related, e.g. `Person`. Keep the error string more indicative of the exception itself so as not to limit it to Person. Maybe, it would relate to animals in the future as well

Dimosphen1 (changes requested):
Several minor changes were requested

Dimosphen1 (approved):
LGTM, well done!

PR #1061, opened 2024-07-21
Implemented solution

dimak20:


app/cafe.py
```diff
@@ -0,0 +1,29 @@
+import datetime
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
```
MNDonut, line 6: import groups should be seprarated by blank line
1. Default python imports
2. Installed imports
3. Your custom imports

app/cafe.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        check = visitor.get("vaccine", False)
```
MNDonut, line 14: No need to set default value False because by default there is None that will not pass if statement below anyway

app/main.py
```diff
@@ -1 +1,23 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            vaccine_check = False
+        except NotWearingMaskError:
+            mask_counter += 1
+    if not vaccine_check:
+        return "All friends should be vaccinated"
+    if mask_counter > 0:
+        return f"Friends should buy {mask_counter} masks"
```
MNDonut, line 22: just if mask_counter is enough

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    vaccine_check = True
+    mask_counter = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            vaccine_check = False
```
MNDonut, line 16: you can return string here immediately to not loop until the last friend

MNDonut (changes requested):

oliasymak24 (approved):

PR #1064, opened 2024-07-25
Solution

DSahalatyi:


app/errors.py
```diff
@@ -0,0 +1,30 @@
+class VaccineError(Exception):
+    def __init__(self) -> None:
+        super().__init__()
```
MNDonut, line 3: no need to add init if you're not going to override it

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    problems_with_vaccines = 0
+    missing_masks = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            problems_with_vaccines += 1
```
MNDonut, line 13: what about Outdated vaccine?
Also it's mentioned that:
`If at least one of the friend has problems with vaccines the function should return a message `"All friends should be vaccinated".`
Thus when you have such a friend, you can immediately return string

MNDonut (changes requested):

oliasymak24 (approved):

PR #1067, opened 2024-07-29
solution

BastovOleksandr:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        visitor_name = visitor["name"]
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(visitor_name)
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError(visitor_name)
+        if "wearing_a_mask" not in visitor or not visitor["wearing_a_mask"]:
```
MNDonut, line 20: you can use .get method

app/main.py
```diff
@@ -1 +1,15 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    return (f"Friends can go to {cafe.name}" if masks_to_buy == 0
```
MNDonut, line 14: let's use `if masks_to_buy` condition

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
```
oliasymak24, line 7: I suggest to separate the loop from two sides with an empty line for better readability

oliasymak24 (approved):

PR #1069, opened 2024-07-30
implement solution

HalynaRohatska:


app/cafe.py
```diff
@@ -0,0 +1,40 @@
...
+    def visit_cafe(self, visitor: dict) -> str:
+        all_check = True
+
+        if not visitor.get("vaccine"):
+            all_check = False
+            raise NotVaccinatedError
+        if date.today() > visitor["vaccine"]["expiration_date"]:
+            all_check = False
+            raise OutdatedVaccineError
+        if not visitor["wearing_a_mask"]:
```
oliasymak24, line 21: in first `if` you use get() function, make it consistent, use it here too

app/main.py
```diff
@@ -1 +1,25 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    checked_friend = True
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            checked_friend = False
+            return "All friends should be vaccinated"
```
oliasymak24, line 16: If at least one of the friend has problems with vaccines the function should return a message "All friends should be vaccinated".`
So, you alredy have there return 
You don't need additionaly check `checked_friend`

oliasymak24 (approved):

PR #1072, opened 2024-08-01
Solution

dartomOOv:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+
+        if visitor["vaccine"].get("expiration_date") < datetime.date.today():
+            raise OutdatedVaccineError("All friends should be vaccinated")
+
+        if not visitor.get("wearing_a_mask"):
+            raise NotWearingMaskError
```
Oleksl888, line 22: It is a bad practice to raise an error without any message

app/errors.py
```diff
@@ -0,0 +1,19 @@
+class VaccineError(Exception):
+    ...
+    # def __str__(self):
```
Oleksl888, line 3: remove unnecessary comments

app/errors.py
```diff
@@ -0,0 +1,19 @@
...
+class NotVaccinatedError(VaccineError):
+    ...
+
+
+class OutdatedVaccineError(VaccineError):
+    ...
+
+
+class NotWearingMaskError(Exception):
+    masks_to_buy = 0
```
Oleksl888, line 16: it is a common practice to pass parameters to `__init__` in exceptions instead of making it a class attribute. Also I see what you did here - you want to track amount of masks inside of the exception but this variable would rather below to go_to_cafe function

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    NotWearingMaskError.masks_to_buy = 0
```
Oleksl888, line 6: it would be better to make `masks_to_buy ` as a normal variable, it will make your code simpler and more obvious to undestand

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError as v_error:
+            return str(v_error)
```
Oleksl888, line 12: As per requirements:
`If at least one of the friend has problems with vaccines the function should return a message `"All friends should be vaccinated".`

app/errors.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def __str__(self) -> str:
```
Oleksl888, line 18: This message doesn't belong here - it should be returned after all friends are processed

app/cafe.py
```diff
@@ -0,0 +1,24 @@
+import datetime
+
+from app.errors import (
+    NotVaccinatedError,
```
Oleksl888, line 4: it is a good practice to sort import alphabetically

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1074, opened 2024-08-01
first try

yashkunn:


app/cafe.py
```diff
@@ -0,0 +1,28 @@
+from datetime import date
+
+from app.errors import (
+    NotVaccinatedError,
```
Oleksl888, line 4: it is a good practice to sort import alphabetically

app/cafe.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("You cannot enter without a vaccine")
+        vaccine_info = visitor["vaccine"]
```
Oleksl888, line 17: you can simplify lines 17-22 to one if condition and raise only one  OutdatedVaccineError

app/cafe.py
```diff
@@ -0,0 +1,28 @@
...
+        expiration_date = vaccine_info.get("expiration_date")
+        if expiration_date is None:
+            raise NotVaccinatedError("Vaccine information is incomplete")
+        if expiration_date < date.today():
+            raise OutdatedVaccineError("The vaccine is expired")
+        if not visitor.get("wearing_a_mask", False):
```
Oleksl888, line 23: it is unnecessary to explicitly set default value of False for .get method - it will return falsy value anyway

app/errors.py
```diff
@@ -0,0 +1,18 @@
+class VaccineError(Exception):
+    """Base class for exceptions related to vaccination."""
```
Oleksl888, line 2: These docstrings provide pretty much same information that is in excpetion name so why do we need them here?

app/main.py
```diff
@@ -1 +1,32 @@
...
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    if not friends:
+        return "No friends to check."
+
+    vaccinated = True
```
Oleksl888, line 13: you can rewrite this function so you don't need to use this variable

app/main.py
```diff
@@ -1 +1,32 @@
...
+    no_mask_count = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotVaccinatedError:
+            vaccinated = False
```
Oleksl888, line 20: Think how you can structure your exceptions so you fulfil this requirement and you don't need to use `vaccinated` variable:

> If at least one of the friend has problems with vaccines the function should return a message `"All friends should be vaccinated".

Oleksl888 (changes requested):

app/errors.py
yashkunn: В теорії було вказано, що докстрінг це гарна практика
+ в усьому, що імпортується(бібліотеки) є докстрінги з уточненнями. 
Якщо в цьому проєкті це заборонено, можу прибрати

Oleksl888: Hey, good point - it is not prohibited to use docstrings. My point was that not all exception classes need docstrings if their name is self-explanatory. For example 
```
NotWearingMaskError = Exception raised when a visitor is not wearing a mask

So it's not really a mistake, in real life it may be project-specific

File: app/main.py [hunk 7]
```diff
@@ -1 +1,27 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            no_mask_count += 1
+
+    if no_mask_count > 0:
```
oliasymak24, line 24: You can just if masks_to_buy

app/cafe.py
oliasymak24: Did you change this?

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1077, opened 2024-08-02
implement lockdown cafe

vladislav-tsybuliak1:


app/cafe.py
```diff
@@ -0,0 +1,28 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
```
LiudmylaKulinchenko, line 15: What if it's none? Use get

app/cafe.py
```diff
@@ -0,0 +1,28 @@
...
+            raise NotVaccinatedError(
+                f"{visitor['name']} is not vaccinated!"
+            )
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError(
+                f"{visitor['name']} has an outdated vaccine!"
+            )
+        if visitor["wearing_a_mask"] is False:
```
LiudmylaKulinchenko, line 23: Let's also cover if there is no wearing_a_mask key (u can use get again)

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
LiudmylaKulinchenko, line 16: You can check if masks_to_buy here

LiudmylaKulinchenko (changes requested):

oliasymak24 (approved):

PR #1078, opened 2024-08-02
Solution

idubina:


app/main.py
```diff
@@ -1 +1,20 @@
...
+            except NotWearingMaskError:
+                masks_to_buy += 1
+        if masks_to_buy == 1:
+            return f"Friends should buy {masks_to_buy} mask"
+        elif masks_to_buy > 1:
+            return f"Friends should buy {masks_to_buy} masks"
+    except VaccineError:
+        return "All friends should be vaccinated"
+    else:
+        return f"Friends can go to {cafe.name}"
```
vsmutok, line 20:
```suggestion
        return "All friends should be vaccinated"
    return f"Friends can go to {cafe.name}"
```
`else` redundant in this case

app/cafe.py
```diff
@@ -0,0 +1,17 @@
+from datetime import date
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError, NotWearingMaskError)
```
vsmutok, line 3:
```suggestion
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError, 
    NotWearingMaskError
)
```

app/main.py
```diff
@@ -1 +1,20 @@
...
+    try:
+        for friend in friends:
+            try:
+                cafe.visit_cafe(friend)
+            except NotWearingMaskError:
+                masks_to_buy += 1
+        if masks_to_buy == 1:
+            return f"Friends should buy {masks_to_buy} mask"
+        elif masks_to_buy > 1:
+            return f"Friends should buy {masks_to_buy} masks"
```
vsmutok, line 16:
```suggestion
        if masks_to_buy > 0:
            return f"Friends should buy {masks_to_buy} mask"
```
I think in this case just one check is enough

vsmutok (changes requested):

app/main.py
idubina: I agree thank you :)

app/main.py
```diff
@@ -1 +1,17 @@
...
+    try:
+        for friend in friends:
+            try:
+                cafe.visit_cafe(friend)
+            except NotWearingMaskError:
+                masks_to_buy += 1
+        if masks_to_buy > 0:
+            return f"Friends should buy {masks_to_buy} masks"
+    except VaccineError:
+        return "All friends should be vaccinated"
```
Dimosphen1, line 16: Try to avoid nesting `try except` blocks, use one `try except` block with multiple `except` statements instead

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
Dimosphen1, line 5: Indicate a list of what instances you expect to obtain in the type annotation in the future

Dimosphen1 (approved):

PR #1082, opened 2024-08-04
Solution

olena4reunited:


app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None:
+        today = datetime.date.today()
+
+        if "vaccine" not in visitor:
```
oliasymak24, line 17: What if it's none? Use get

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+            raise NotVaccinatedError("Visitors must be vaccinated!")
+        elif visitor["vaccine"]["expiration_date"] < today:
+            raise OutdatedVaccineError(
+                "The vaccine is outdated! Visitor need to renew vaccination!"
+            )
+        elif not visitor["wearing_a_mask"]:
```
oliasymak24, line 23: All of these `elif` must be separate `ifs`

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (VaccineError):
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
oliasymak24, line 16: you can just if masks_to_buy

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1083, opened 2024-08-04
implement py-lockdown-cafe

ValeriiaKyr:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+    OutdatedVaccineError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
```
Oleksl888, line 15: Can you rewrite this for efficiency? Hint: in dictionary finding an element by a key has the best possible algorythmic coplexity

app/main.py
```diff
@@ -1 +1,17 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            Cafe.visit_cafe(cafe, friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy > 0:
```
Oleksl888, line 14: Can be simply `if masks_to_buy` here

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("The visitor must have a vaccine")
+        if visitor.get("vaccine").get("expiration_date") < date.today():
```
MNDonut, line 17: If vaccine key exists, but it's None, then you'll try to do None.get('expiration_date') that will cause an error. When you develop something like this, you should try to handler such cases because you never know what data is used in tests or what data you'll get when a real user runs your program

MNDonut (changes requested):

app/cafe.py
```diff
@@ -12,9 +12,10 @@ def __init__(self, name: str) -> None:
         self.name = name
 
     def visit_cafe(self, visitor: dict) -> str:
-        if not visitor.get("vaccine"):
+        if not visitor.get("vaccine") or visitor.get("vaccine") is None:
```
MNDonut, line 15: you can do it in 1 condition using "not" ;)

MNDonut (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+    OutdatedVaccineError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if (not visitor.get("vaccine")
```
Oleksl888, line 15: `if (not visitor.get("vaccine")` is enough here it - read the docs about  what does `.get` return

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+                and not None):
+            raise NotVaccinatedError("The visitor must have a vaccine")
+        if (visitor.get("vaccine").get("expiration_date") < date.today()
+                and not None):
```
Oleksl888, line 19: why do you need ` and not None` condition here?

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+            raise OutdatedVaccineError("The vaccine must not be expired")
+        if not visitor.get("wearing_a_mask", False):
```
Oleksl888, line 21: `.get` method already returns falsy default value

Oleksl888 (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("The visitor must have a vaccine")
+        if (visitor.get("vaccine").get("expiration_date") < date.today()
+                or visitor.get("vaccine").get("expiration_date") is None):
```
Oleksl888, line 18: This condition should come first because if expiration_date is None it will cause ValueError on previous line (you cannot compare Nonetype to datetime). Also you don't need to check that it `is None` explicitly

PR #1085, opened 2024-08-04
first

panichqa:


app/cafe.py
```diff
@@ -0,0 +1,28 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None:
+        if "vaccine" not in visitor:
```
oliasymak24, line 14: What if it's none? Use get

app/main.py
```diff
@@ -1 +1,25 @@
...
+        except (NotVaccinatedError, OutdatedVaccineError):
+            all_vaccinated = False
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if not all_vaccinated:
+        return "All friends should be vaccinated"
+    elif masks_to_buy > 0:
+        return f"Friends should buy {masks_to_buy} masks"
+    else:
```
oliasymak24, line 24: Unnecessary `else`, you can call return directly after `if` statement

app/main.py
```diff
@@ -1 +1,25 @@
...
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(friends: list, cafe: str) -> None:
+    masks_to_buy = 0
+    all_vaccinated = True
```
oliasymak24, line 10: You don't need this flag

app/main.py
```diff
@@ -1 +1,25 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            all_vaccinated = False
```
oliasymak24, line 16: You don't need `all_vaccinated` variable, try to rewrite it
> If at least one of the friend has problems with vaccines the function should return a message `"All friends should be vaccinated".

oliasymak24 (changes requested):

oliasymak24 (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,28 @@
+from datetime import date
```
Oleksl888, line 1: different import types should be separated by blank line

app/cafe.py
```diff
@@ -0,0 +1,28 @@
...
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError(f"{visitor['name']} is not vaccinated.")
+
+        expiration_date = visitor["vaccine"].get("expiration_date")
+        if expiration_date < date.today():
+            raise OutdatedVaccineError(
+                f"{visitor['name']}'s vaccine is expired."
+            )
+
+        if not visitor.get("wearing_a_mask", False):
```
Oleksl888, line 23: no need to explicitly set False as default value

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(friends: list, cafe: str) -> str:
```
Oleksl888, line 8: cafe parameter has wrong annotation

app/main.py
```diff
@@ -1 +1,22 @@
...
+    masks_to_buy = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
Oleksl888, line 14: What you did is correct - but in task requirements it is asking to catch different type of error

Oleksl888 (changes requested):

MNDonut (approved):

PR #1086, opened 2024-08-04
Solution

Bogdan-Rylov:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
```
oliasymak24, line 13: What if it's none? Use get

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    unvaccinated_friends_count = 0
+    lacking_masks_count = 0
```
oliasymak24, line 7: You don't need this flag

app/main.py
```diff
@@ -1 +1,22 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotWearingMaskError:
+            lacking_masks_count += 1
+        except VaccineError:
+            unvaccinated_friends_count += 1
+
+    if unvaccinated_friends_count != 0:
+        return "All friends should be vaccinated"
```
oliasymak24, line 18: If at least one of the friend has problems with vaccines the function should return a message `"All friends should be vaccinated".
you don't need to count all of them

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    lacking_masks_count = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotWearingMaskError:
+            lacking_masks_count += 1
+        except VaccineError:
```
Oleksl888, line 13: This exception should be raised first

app/main.py
```diff
@@ -1 +1,19 @@
...
+            return "All friends should be vaccinated"
+
+    if lacking_masks_count != 0:
```
Oleksl888, line 16: no need to explicitly compare to 0 here

Oleksl888 (changes requested):

MNDonut (approved):

PR #1087, opened 2024-08-04
Solution of py-lockdown-cafe

AndriiHamasa:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | None:
+        if not ("vaccine" in visitor):
```
oliasymak24, line 15: What if it's none? Use get

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
```
oliasymak24, line 7: Add empty line before loop

oliasymak24 (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | None:
+        if visitor.get("vaccine") is None:
+            raise NotVaccinatedError("NotVaccinatedError")
```
MNDonut, line 16: if there is vaccine but it's an empty dict, then the condition below will raise a key error, let's handle this case as well

app/main.py
```diff
@@ -1 +1,19 @@
...
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy != 0:
+        return f"Friends should buy {masks_to_buy} masks"
+
+    return f"Friends can go to {cafe.name}"
```
MNDonut, line 19: let's use a backward condition

MNDonut (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str | None:
```
Oleksl888, line 14: Use correct return type. If your method raises an exception it doesn't mean it will return None

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+        if (
+                visitor.get("vaccine") is None
```
Oleksl888, line 16: Here `if not visitor.get("vaccine")` would be enough to check if visitor is vaccinated

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+                or len(visitor.get("vaccine").items()) == 0
+        ):
+            raise NotVaccinatedError("NotVaccinatedError")
```
Oleksl888, line 19: Error messages must be more descriptive - not just copying exception name

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy == 0:
```
Oleksl888, line 16: what would be more `pythonic` way to evaluate this?

Oleksl888 (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,30 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if visitor.get("vaccine") is None:
```
oliasymak24, line 15: if not visitor.get("vaccine") would be enough to check if visitor is vaccinated

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy != 0:
```
oliasymak24, line 16: `if masks_to_buy` will be enough here

oliasymak24 (changes requested):

PR #1088, opened 2024-08-04
create cafe with check mask adn vaccine

egor250330:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
```
oliasymak24, line 12: What if it's none? Use get

app/main.py
```diff
@@ -1 +1,21 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            mask_to_buy += 1
+
+    if mask_to_buy == 0:
+        return f"Friends can go to {cafe.name}"
+    else:
```
oliasymak24, line 20: This is unnecessary `else`, you can use return directly after `if` statement

oliasymak24 (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,22 @@
+import datetime
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
MNDonut, line 4: you can import a few objects from a package together

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if visitor.get("vaccine") is None:
+            raise NotVaccinatedError("The visitor doesn't have a vaccine!")
```
MNDonut, line 13: what if vaccine is an empty dict? The error will not be raised but the line below will raise a key error

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
MNDonut, line 4: the same here

MNDonut (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,26 @@
+import datetime
```
Oleksl888, line 1: it is a good practice to separate different types of imports with blank lines + arranging them alphabetically

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        vaccine_information = visitor.get("vaccine")
```
Oleksl888, line 14: no need to store this in variable since you are not going to reuse it

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+
+        if not vaccine_information:
+            raise NotVaccinatedError("The visitor doesn't have a vaccine")
+
+        expiration_date = visitor["vaccine"]["expiration_date"]
```
Oleksl888, line 19: same as above

app/main.py
```diff
@@ -1 +1,22 @@
...
+)
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    mask_to_buy = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
Oleksl888, line 15: Fix as per task requirements:

> Create a VaccineError parent class for NotVaccinatedError and OutdatedVaccineError errors. Use it to catch both types of errors in the same except clause.

app/main.py
```diff
@@ -1 +1,22 @@
...
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            mask_to_buy += 1
+
+    if mask_to_buy == 0:
```
Oleksl888, line 20: `if not mask_to_buy` should be enough to evaluate this expression. It is more `pythonic` way :wink:

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            mask_to_buy += 1
+
+    if mask_to_buy == 0:
```
oliasymak24, line 19: Change this `if`, no need to put `mask_to_buy == 0`, zero is False

oliasymak24 (changes requested):

app/cafe.py
Oleksl888: Hasn't been done yet

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("The visitor doesn't have a vaccine")
+
+        elif visitor["vaccine"].get("expiration_date") < datetime.date.today():
+            raise OutdatedVaccineError("The visitor's vaccine is outdated!")
+
+        if not visitor.get("wearing_a_mask", False):
```
Oleksl888, line 21: No need to set default value of False - it will return None otherwise which will produce same result in this scenario

app/main.py
```diff
@@ -1 +1,21 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            mask_to_buy += 1
+
+    if mask_to_buy > 0:
```
Oleksl888, line 19: no need to compare to 0 explicitly here

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #1089, opened 2024-08-04
Task Solution

achalaidiuk:


app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+
+        if "vaccine" not in visitor:
```
oliasymak24, line 16: What if it's none? Use get

oliasymak24 (changes requested):

MNDonut (approved):

PR #1092, opened 2024-08-05
'Solution'

ABaidin:


app/main.py
```diff
@@ -1 +1,19 @@
...
+    for person in friends:
+        try:
+            cafe.visit_cafe(person)
+
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
Oleksl888, line 16: There is a more pythonic way to make this expression work

Oleksl888 (changes requested):

app/main.py
ABaidin: I guess you mean 
"if masks_to_buy:"

app/main.py
```diff
@@ -1 +1,19 @@
...
+from app.cafe import Cafe
+from app.errors import NotWearingMaskError, VaccineError
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for person in friends:
+        try:
+            cafe.visit_cafe(person)
+
```
oliasymak24, line 10: Why do you this emty line?

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import NotWearingMaskError, VaccineError
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for person in friends:
```
oliasymak24, line 7: Add an empty line before the loop

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1093, opened 2024-08-05
Solution

Ivan-Shakhman:
Solution of task

### DIFF
--- a/app/cafe.py
+++ b/app/cafe.py
@@ -1,0 +1,23 @@
+import datetime
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
+            raise NotVaccinatedError("Visitor is not vaccinated")
+        if (
+            visitor["vaccine"]["expiration_date"] is None
+            or visitor["vaccine"]["expiration_date"] < datetime.date.today()
+        ):
+            raise OutdatedVaccineError("Vaccine of visitor is expired")
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("Visitor must have a mask")
+        return f"Welcome to {self.name}"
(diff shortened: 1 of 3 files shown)


app/cafe.py
```diff
@@ -0,0 +1,23 @@
+import datetime
```

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
```
Oleksl888, line 14: Think how you can evaluate this more efficiently. Hint: accessing a value in dictionary by key gives you best algorythmic complexity

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+            raise NotVaccinatedError("Visitor is not vaccinated")
+        if (
+            visitor["vaccine"]["expiration_date"] is None
```
Oleksl888, line 17: you can rewrite this expression in one line

app/main.py
```diff
@@ -1 +1,23 @@
...
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
Oleksl888, line 15: You should raise different type of error here as per task requirements

app/main.py
```diff
@@ -1 +1,23 @@
...
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+            continue
```
Oleksl888, line 19: why do use this instruction here ?

app/main.py
```diff
@@ -1 +1,23 @@
...
+    return (
+        f"Friends can go to {cafe.name}" if masks_to_buy == 0
```
Oleksl888, line 21: you don't need to compare  masks_to_buy to zero explicitly here

Oleksl888 (changes requested):

PR #1095, opened 2024-08-05
Solution

pashawarganov:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor.keys():
```
Oleksl888, line 16: think how you can optmize this line for the best performance. Hint: getting an element from dictionary by key is the fastest operation

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
```
Oleksl888, line 1: it is a good practice to arrange imports alphabetically

app/main.py
```diff
@@ -1 +1,17 @@
...
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    mask_to_buy = 0
+    try:
```
Oleksl888, line 7: you can rewrite this so you have only one `try` block

Oleksl888 (changes requested):

app/main.py
pashawarganov: It was mentioned in the lectures and it was said that it is not mandatory. I think this practice will not improve readability.

oliasymak24 (approved):

PR #1096, opened 2024-08-05
task done

Sergi0bbb:


app/cafe.py
```diff
@@ -0,0 +1,23 @@
+import datetime
```

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
```
Oleksl888, line 14: Think how you can evaluate this more efficiently. Hint: accessing a value in dictionary by key gives you best algorithmic complexity

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+            raise NotVaccinatedError("Visitor is not vaccinated.")
+
+        if visitor["vaccine"].get("expiration_date") < datetime.date.today():
+            raise OutdatedVaccineError("Vaccine has expired.")
+
+        if visitor.get("wearing_a_mask") is False:
```
Oleksl888, line 20: `.get` method returns `None` as a default value so comparing to `False` is not correct. How can you simplify this condition and make it more pythonic?

app/main.py
```diff
@@ -1 +1,27 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            vaccine = False
+        except NotWearingMaskError:
+            masks_needed += 1
+
+    if vaccine is False:
+        return "All friends should be vaccinated"
```
Oleksl888, line 22: Could you return this in `except` clause? Then you won't need `vaccine` variable at all

app/main.py
```diff
@@ -1 +1,27 @@
...
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    vaccine = True
+    masks_needed = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
Oleksl888, line 16: There should be a different error raised here as per task requirements

app/main.py
```diff
@@ -1 +1,27 @@
...
+            vaccine = False
+        except NotWearingMaskError:
+            masks_needed += 1
+
+    if vaccine is False:
+        return "All friends should be vaccinated"
+
+    if masks_needed > 0:
```
Oleksl888, line 24: You can rewrite this condition slightly to make it more pythonic

Oleksl888 (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,23 @@
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+import datetime
```
Oleksl888, line 6: Check order of imports + is a good practice to separate different kinds of imports with blank lines

app/cafe.py
Oleksl888: when you use `not in` it means that you iterate through all the keys in dictionary until you find a match. When you access an element by a key it is instant match. Please change

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated.")
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Vaccine has expired.")
+
+        if not visitor["wearing_a_mask"]:
```
Oleksl888, line 20: would be better to use .get method here in case if this key is not in dictionary

app/main.py
Oleksl888: do you really need to compare masks_needed to 0?

Oleksl888 (changes requested):

app/cafe.py
Oleksl888: please change

Oleksl888 (changes requested):

app/cafe.py
Sergi0bbb: I've already changed it. Maybe you opened the wrong commit?

Oleksl888: No you haven't

Oleksl888 (approved):
Approved but please fix the last comment

PR #1098, opened 2024-08-05
Solution

mnezbrytska:


app/cafe.py
```diff
@@ -0,0 +1,27 @@
+import datetime
```

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+)
+
+
+class Cafe:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        all_rule_done = True
```
Oleksl888, line 15: You don't need this variable to make it work - simply return result on last line if none of the `if` conditions trigger an exception

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+        if visitor.get("vaccine") is None:
```
Oleksl888, line 16: `if not visitor.get("vaccine")` will be more `pythonic` way for this expression

app/main.py
```diff
@@ -1 +1,20 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
Oleksl888, line 17: `if not masks_to_buy` will be more pythonic way for this expression

Oleksl888 (changes requested):

app/cafe.py
Oleksl888: Remove all_rule_done variable and refactor code below so it works without it

Oleksl888 (changes requested):

Oleksl888 (approved):
Good job!

PR #1099, opened 2024-08-05
Solution

Poznanskyi:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None:
+        if "vaccine" not in visitor:
```
oliasymak24, line 12: What if it's none? Use get function

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> None:
```
oliasymak24, line 11: Are you sure that this function return None?

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated.")
+
+        vaccine_info = visitor["vaccine"]
```
oliasymak24, line 15: No need to add this additional variable

app/main.py
```diff
@@ -1 +1,17 @@
...
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
oliasymak24, line 14: just `if masks_to_buy` will be enough

app/main.py
```diff
@@ -1 +1,17 @@
...
+        return f"Friends should buy {masks_to_buy} masks"
+    else:
```
oliasymak24, line 16: You don't need this `else`, just use return after `if` statement

oliasymak24 (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,21 @@
+import datetime
```
Oleksl888, line 1: Different import types should be separated by blank line
Also it is a common practice to list imports from same module separated by commas

app/cafe.py
Oleksl888: Using `.get` would be more efficient here

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated.")
+
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Vaccine is outdated.")
+
+        if not visitor.get("wearing_a_mask", False):
```
Oleksl888, line 18: no need to explicitly set default value to False since it returns None already

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: dict, cafe: str) -> str:
```
Oleksl888, line 4: use correct type annotation on cafe

Oleksl888 (changes requested):

Oleksl888 (approved):
Good job!

PR #1101, opened 2024-08-06
Resolve

ihorutkin:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
+import datetime
```
Oleksl888, line 1: There must be a blank line between different types of imports. 
Imports from same module can be listed separated by comma instead

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from __future__ import annotations
```
Oleksl888, line 1: Is this import necessary? everything should work without it. But if you want to keep it you must add blank line between. Also list imports from app.errors in one row

app/main.py
```diff
@@ -1 +1,20 @@
...
+    masks_to_buy = 0
+    for person in friends:
+        try:
+            cafe.visit_cafe(person)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
Oleksl888, line 17: no need to explicitly compare to zero - if it's NOT a falsy value expression will return TRUE

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1104, opened 2024-08-08
Develop

alheimurin-n:


.github/workflows/test.yml
```diff
@@ -27,5 +27,5 @@ jobs:
         run: flake8 app/
       - name: Run tests
         timeout-minutes: 5
-        run: pytest tests/
+        run: pytest
```
NazarSheremet, line 30: Why change the test run directory?

NazarSheremet (changes requested):
A fix or explanation is needed

.github/workflows/test.yml
alheimurin-n: It was a temporary solution that I forgot to remove

NazarSheremet (approved):
Good job!

PR #1112, opened 2024-08-16
My solution

lavryniuk-dev:


LiudmylaKulinchenko (approved):

app/main.py
```diff
@@ -1 +1,21 @@
...
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            without_masks += 1
+
+    if without_masks > 0:
+        return f"Friends should buy {without_masks} masks"
+    else:
+        return f"Friends can go to {cafe.name}"
```
LiudmylaKulinchenko, line 21: No need to use else if you have return in if statement

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌼

PR #1137, opened 2024-09-04
Solution

FalseCoreCentury:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
```
LLkaia, line 16:
```suggestion
        vaccine = visitor.get("vaccine")
        if not vaccine:
            raise NotVaccinatedError("Visitor is not vaccinated")
```
And then you can reuse `vaccine` variable

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+
+        expiration_date = visitor["vaccine"].get("expiration_date")
```
LLkaia, line 18: e.g. reuse `vaccine` here

LLkaia (changes requested):

LLkaia (approved):
Perfect

PR #1149, opened 2024-09-12
Solution

CheshireKate:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
```
Dimosphen1, line 1: Add a blank line between the different import groups to follow PEP-8

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError, NotWearingMaskError)
```
Dimosphen1, line 3: Place each import on a new line so there would not be imports on the same line with opening or closing parenthesis. It would help to improve the readability

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+from app.cafe import Cafe
```
Dimosphen1, line 2: Order your imports in alphabetical order

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    all_vaccinated = True
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            all_vaccinated = False
```
Dimosphen1, line 12: There is no need for the `all_vaccinated` flag, in case even one person is not vaccinated you could return the required string straight away. Remove this flag

Dimosphen1 (changes requested):

PR #1152, opened 2024-09-16
solution for lockdown cafe

dissom:


app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        vaccinated = visitor.get("vaccine", None)
+        mask = visitor.get("wearing_a_mask", None)
```
YukihiroSM, line 16:
```suggestion
        vaccinated = visitor.get("vaccine")
        mask = visitor.get("wearing_a_mask")
```

Get returns None by default

YukihiroSM (approved):
Anything else looks pretty good to me, so there is an approve. But just note, that get() method can be used without "None" as parameter

app/cafe.py
dissom: I know,but in vscode i have autoinput and i forgot to check in this place)))

PR #1166, opened 2024-09-30
Solution

KKvladys:


app/main.py
```diff
@@ -1 +1,17 @@
...
+    count_not_wear_mask = 0
+    for person in friends:
+        try:
+            cafe.visit_cafe(person)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            count_not_wear_mask += 1
+
+    if count_not_wear_mask > 0:
```
Atikiho, line 15: You don't need to compare with 0 here, just use 
`if count_not_wear_mask:` as it was told in checklist

Atikiho (changes requested):

PR #1168, opened 2024-09-30
Solution

Atikiho:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list[dict], cafe: "Cafe") -> str:
```
AnyoneClown, line 5: Don't forget, that u can import annotation from `__future__` to use class as annotation without quotes

AnyoneClown (approved):
Good job!

PR #1171, opened 2024-10-04
Solution_1

apmaz:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
+import datetime
+
+
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
Dimosphen1, line 6: Do not duplicate imports, use `from app.errors import` once and place all required imports inside of `()` so that there would be no imports on the same line with opening and closing parenthesis

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError
+from app.errors import NotWearingMaskError
```
Dimosphen1, line 3: Import two errors on a one line

pavlopro (approved):
GJ!

PR #1175, opened 2024-10-06
Solution

vladyslav-tmf:


app/main.py
```diff
@@ -1 +1,18 @@
...
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
Mxbely, line 16: you can use just `if masks_to_buy:`

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated.")
+
+        expiration_date = visitor["vaccine"].get("expiration_date")
+        if expiration_date < datetime.date.today():
+            raise OutdatedVaccineError("Visitor's vaccine is outdated.")
+
+        if not visitor.get("wearing_a_mask", False):
+            raise NotWearingMaskError("Visitor is not wearing a mask.")
```
Mxbely, line 23: if I know, when we raises errors, we need output name errors
`raise OutdatedVaccineError("OutdatedVaccineError")`
image

Mxbely (approved):
You code is ok, but there is one momont... (see comments)

app/main.py
vladyslav-tmf: Fixed.

app/cafe.py
vladyslav-tmf: I'm not sure about that. In case of an error, I get the name of the error and its description in the terminal.

AnyoneClown: This is good, we need text of error, to explain user why did this error raised(His solution is good)

AnyoneClown (approved):
LGTM!

PR #1177, opened 2024-10-07
Solution

dmytrostepanenko:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+                "The visitor does not have a vaccine key."
+            )
+        if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError(
+                "The vaccine must not be expired"
+            )
+        if "wearing_a_mask" not in visitor.keys() or \
+                visitor["wearing_a_mask"] is False:
+            raise NotWearingMaskError(
+                "All visitors must wear masks"
```
vsmutok, line 23: better to use `()` then `\`

app/cafe.py
```diff
@@ -0,0 +1,25 @@
+import datetime
+
+from app.errors import (NotVaccinatedError, NotWearingMaskError,
+                        OutdatedVaccineError)
```
```suggestion
from app.errors import (
    NotVaccinatedError, 
    NotWearingMaskError,
    OutdatedVaccineError
)
```

vsmutok (changes requested):

app/cafe.py
dmytrostepanenko: ) First I used (), then chenged to \
Thank you!

app/cafe.py
dmytrostepanenko: And here...
First I wrote imports to column, but "isort" changed my code )

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import NotWearingMaskError, VaccineError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
```
```suggestion
def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1178, opened 2024-10-07
lockdown cafe solution

94nj111:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+
+from app.errors import NotVaccinatedError
+from app.errors import NotWearingMaskError
+from app.errors import OutdatedVaccineError
```
vladyslav-tmf, line 5: This import is correct, but it is a good practice to use the following variation:
```suggestion
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError,
)
```

app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("All friends should be vaccinated")
+        if visitor["vaccine"].get("expiration_date") < datetime.date.today():
+            raise OutdatedVaccineError("Your vaccine is out of date")
+        if not visitor.get("wearing_a_mask"):
+            raise NotWearingMaskError("You should wear mask")
+        else:
+            return f"Welcome to {self.name}"
```
vladyslav-tmf, line 20: You can use `return` without `else`.

vladyslav-tmf (changes requested):

vladyslav-tmf (approved):
Good job)

app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+                        NotWearingMaskError,
+                        OutdatedVaccineError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
```
AnyoneClown, line 13: Good, but read about warlus operator. We can use it here

[Click](https://realpython.com/python-walrus-operator/)

AnyoneClown (approved):
Good job! Check comment

PR #1200, opened 2024-10-14
Solution

molodsh1y1:


skyfoxwork (approved):
Good job!

app/main.py
```diff
@@ -1 +1,22 @@
...
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
skyfoxwork, line 19: You can use:

        if masks_to_buy:
            return f"Friends should buy {masks_to_buy} masks"
            
instead of 

    if masks_to_buy > 0:
        return f"Friends should buy {masks_to_buy} masks"

molodsh1y1: thx I miss about it ;)

PR #1209, opened 2024-10-15
solution

YaYaYaroslav:


app/main.py
```diff
@@ -1 +1,25 @@
...
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except NotVaccinatedError:
```
Le0n-K, line 15: here you can use parent class (VaccineError) instead of using 2 child class, and make your code less :)

Le0n-K (approved):
looks good)

PR #1210, opened 2024-10-15
Solution

YaroslavK87:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("The visitor should be vaccinated.")
+        vaccine_expiration_date = visitor["vaccine"]["expiration_date"]
+        today = datetime.date.today()
+        if vaccine_expiration_date < today:
```
Rabb1t-Artem, line 18: do not create additional variables, you can check this in the "if" construct.

app/errors.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class OutdatedVaccineError(VaccineError):
+    def __init__(self, message: str) -> None:
+        self.message = message
+
+
+class NotWearingMaskError(Exception):
+    def __init__(self, message: str) -> None:
+        self.message = message
```
Rabb1t-Artem, line 17: you have init methods, but don't use it anywhere

app/main.py
```diff
@@ -1 +1,17 @@
...
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy > 0:
```
Rabb1t-Artem, line 14: if you want, you can change "if masks_to_buy > 0:" to "if masks_to_buy :" it will be same.

Rabb1t-Artem (changes requested):

PR #1278, opened 2024-12-09
created all packages clases and methods

shafran-andriy:


app/cafe.py
```diff
@@ -0,0 +1,27 @@
+import datetime
+
+from app.errors import NotVaccinatedError
```
pavlopro: No need to have import lines. Combine into one

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+        if not vaccine:
+            raise NotVaccinatedError(
+                "All friends should be vaccinated")
+        if (not expiration_date
+                or expiration_date < datetime.date.today()):
+            raise OutdatedVaccineError(
+                "All friends should be vaccinated")
+        if not visitor.get("wearing_a_mask", False):
+            raise NotWearingMaskError(
+                "Friends should buy {masks_to_buy} masks")
```
pavlopro, line 27: what is masks_to_buy ?

pavlopro (changes requested):

app/cafe.py
shafran-andriy: it is too long (more than 79) if I create one line That why i import classes in 3 lines
Or is it a mistake?

app/cafe.py
shafran-andriy: masks_to_buy - is the count of friends who do not wear the mask

app/cafe.py
```diff
@@ -0,0 +1,27 @@
+import datetime
+
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
+from app.errors import NotWearingMaskError
```
pavlopro:
```
from app.errors import (
    NotVaccinatedError,
    NotWearingMaskError,
    OutdatedVaccineError
)
```

app/cafe.py
pavlopro: Once more 
 "Friends should buy {masks_to_buy} masks"
 What is the masks_to_buy here ???

pavlopro (changes requested):

app/cafe.py
shafran-andriy: fixed

shafran-andriy:
masks_to_buy is the number of friends who don't have a mask

vsmutok (approved):
Great!

PR #1295, opened 2024-12-14
Solution

VovaGld:


app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+                        NotWearingMaskError,
+                        NotVaccinatedError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        no_exception = True
```
taraskhom: You can remove this variable and not use it, because when an error is called your code will stop going any further, so it's unnecessary code

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+        if "vaccine" in visitor:
+            if visitor["vaccine"]["expiration_date"] < datetime.date.today():
+                no_exception = False
```
taraskhom: unnecessary code

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+                raise OutdatedVaccineError("OutdatedVaccineError")
+        else:
+            no_exception = False
```
taraskhom: unnecessary code

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+            raise NotVaccinatedError("NotVaccinatedError")
+
+        if not visitor["wearing_a_mask"]:
+            no_exception = False
```
taraskhom: unnecessary code

app/cafe.py
```diff
@@ -0,0 +1,27 @@
...
+            raise NotWearingMaskError("NotWearingMaskError")
+
+        if no_exception:
```
taraskhom: No validation needed, you can just return the value

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.errors import VaccineError, NotWearingMaskError
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    masks_to_buy = 0
+    no_exception = True
```
taraskhom: You can remove this variable and not use it

app/main.py
```diff
@@ -1 +1,20 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            no_exception = False
```
taraskhom: unnecessary code

app/main.py
```diff
@@ -1 +1,20 @@
...
+            masks_to_buy += 1
+
+    if no_exception:
```
taraskhom: You can simply check if there are masks for purchase

taraskhom (changes requested):

taraskhom (approved):

PR #1301, opened 2024-12-15
Solution

Chukhno-Maksym:


app/main.py
```diff
@@ -1 +1,22 @@
...
+        try:
+            cafe.visit_cafe(person)
+
+        except VaccineError:
+            return "All friends should be vaccinated"
+
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy > 0:
```
taraskhom, line 19: You can simplify to " if masks_to_buy: " because if the value is zero it will automatically be False

taraskhom (approved):
Good job!

PR #1302, opened 2024-12-15
solution

DanilTrukhanov:


app/cafe.py
```diff
@@ -0,0 +1,22 @@
+import datetime
```
ruslanyemets, line 1: Standard library import and local application specific imports should be separated by single blank line

app/main.py
```diff
@@ -1 +1,19 @@
...
+    masks_to_buy = 0
+    for person in friends:
+        try:
+            cafe.visit_cafe(person)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy != 0:
```
ruslanyemets: You can use condition without comparing a variable to 0

ruslanyemets (approved):

app/cafe.py
petrykivd: I agree with Ruslan

app/main.py
```diff
@@ -1 +1,19 @@
...
+        return f"Friends should buy {masks_to_buy} masks"
+    else:
+        return f"Friends can go to {cafe.name}"
```
petrykivd: Please do not forget to read the checklist before completing the task. It says about else and return.

petrykivd (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+from app.errors import NotWearingMaskError
+from app.errors import NotVaccinatedError
+from app.errors import OutdatedVaccineError
```
L1nk3rrr: single `from` import multiple classes, read a checklist

app/cafe.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("You are not vaccinated!")
+
+        today = datetime.date.today()
```
L1nk3rrr: do you need an extra variable here?

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError
+from app.errors import NotWearingMaskError
```
L1nk3rrr: same about imports

L1nk3rrr (changes requested):
Great work! Added extra point for improvement of your code

PR #1317, opened 2024-12-16
Task

Arty-tech-dotcom:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+import datetime
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
```
MaksVakulenko, line 6: maybe it's better just from app import errors

MaksVakulenko (approved):
very strange imports

app/cafe.py
petrykivd: There should be an empty line between the imports of default libraries and custom files.

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitors: dict) -> str:
+        if "vaccine" not in visitors.keys():
```
petrykivd, line 14: Are you sure you need .keys() here?

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    bed_result = []
+    neg_count = 0
```
petrykivd, line 7: Why do you need an extra list? Isn't just counter enough?

petrykivd (changes requested):

PR #1318, opened 2024-12-16
solution to lockdown cafe

SummerKatrin:


bodiakof (approved):
Nicely done!

app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(f'{visitor["name"]} is not vaccinated.')
+        if ("expiration_date" not in visitor["vaccine"]
+                or visitor["vaccine"]["expiration_date"]
+                < datetime.date.today()):
+            raise OutdatedVaccineError(
```
petrykivd, line 20: Let's simplify, if visitor already have "vaccine" key, in this case, it should definitely have an expiration date. 
Simplify in another places too

petrykivd (changes requested):

PR #1323, opened 2024-12-17
Solution

shretsker:


app/errors.py
```diff
@@ -0,0 +1,16 @@
+class VaccineError(Exception):
+    def __init__(self, message: str) -> None:
```
yelizaveta-zh: You raise all exceptions at cafe.py. You don't have to rewrite the same thing in this file, it's enough to write "pass" for each class here.

yelizaveta-zh (approved):

app/errors.py
petrykivd: Agree with Yelizaveta

app/main.py
```diff
@@ -1 +1,21 @@
...
+    NotWearingMaskError
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(visitor=friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
petrykivd: You must except here "Vaccine Error".  NotVaccinatedError, OutdatedVaccineError Inherited from VaccineError.

app/main.py
```diff
@@ -1 +1,21 @@
...
+            return "All friends should be vaccinated"
```
petrykivd: Don't duplicate this string here and in cafe class. You can return here text of error. (as e)

petrykivd (changes requested):

petrykivd (approved):
Nice

PR #1354, opened 2025-01-06
Solution

Baranotik15:


app/__init__.py
oliasymak24, line 1: You don't need to push this file

app/cafe.py
```diff
@@ -0,0 +1,31 @@
...
+            raise OutdatedVaccineError("Vaccine expiration date is missing.")
+
+        if expiration_date < datetime.date.today():
+            raise OutdatedVaccineError(f"Vaccine expired on"
+                                       f" {expiration_date}.")
+
+        if "wearing_a_mask" not in visitor:
+            raise NotWearingMaskError("Visitor need to wear mask")
+
+        if visitor["wearing_a_mask"] is False:
```
oliasymak24: Use `not` instead of `is False`

oliasymak24 (changes requested):

.gitignore
```diff
@@ -5,4 +5,5 @@
 .DS_Store
 venv/
 .pytest_cache/
-**__pycache__/
+**/__pycache__/
+**/__init__.py
```
Y-Havryliv, line 9: it's a good practice to add an empty line at the end of the files
https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file

Y-Havryliv (approved):

PR #1412, opened 2025-02-12
Solution

Revko:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
+import datetime
```
SerhiyTo, line 1: Please add an indent between the import of the build-in module and the local one

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from __future__ import annotations
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    masks_to_buy = 0
```
SerhiyTo, line 8: You can simplify this code by using the sum function. In this case, you will not need to create separate counter variables

SerhiyTo (changes requested):

app/main.py
Revko: Unfortunately, I can't find a solution using the sum function. Could you give me a hint, please?

vasylhnatiuk (approved):

PR #1417, opened 2025-02-17
solution py-lockdown-cafe

dmytrominyaylo:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_needed = 0
```
SerhiyTo: Simplify this logic. You can use the `sum` function instead of an additional variable.

SerhiyTo (changes requested):

app/main.py
```diff
@@ -1 +1,25 @@
...
+
+
+def try_visit_cafe(friend: Any, cafe: Any) -> Any:
+    try:
+        cafe.visit_cafe(friend)
+        return None
+    except VaccineError:
+        return "vaccine"
+    except NotWearingMaskError:
+        return "mask"
```
vsmutok:
```suggestion
def go_to_cafe(friends: list, cafe: Cafe) -> str:
    masks = 0
    for friend in friends:
        try:
            cafe.visit_cafe(friend)
        except VaccineError:
            return "All friends should be vaccinated"
        except NotWearingMaskError:
            masks += 1
    if masks > 0:
        return f"Friends should buy {masks} masks"
    return f"Friends can go to {cafe.name}"
```

app/errors.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class OutdatedVaccineError(VaccineError):
+    def __init__(self, message: str = "Vaccine is outdated.") -> None:
+        super().__init__(message)
+
+
+class NotWearingMaskError(Exception):
+    def __init__(self, message: str = "Visitor is not wearing a mask.") \
+            -> None:
```
vsmutok: better to use `()` then `\`

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1428, opened 2025-02-23
solution

StenSOn27:


StenSOn27:
я запушив 4 файли чому ж missed 

    Missing errors.py file: The task requires defining all errors in a separate errors.py module, including a VaccineError parent class for NotVaccinatedError and OutdatedVaccineError errors. This file is currently missing and needs to be added to meet the task requirements.

    Implementation of Cafe class and visit_cafe method: The cafe.py file, which should contain the implementation of the Cafe class and the visit_cafe method, is also missing. This is crucial for the functionality of your solution and must be implemented according to the task requirements.

PR #1450, opened 2025-03-02
Solution

Oneyura:


app/main.py
```diff
@@ -1 +1,20 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy:
+        return f"Friends should buy {masks_to_buy} masks"
+    else:
+        return "Friends can go to KFC"
```
Anollinna, line 20: It would be more correct to write f"Friends can go to {cafe.name}"

Anollinna (approved):
Great job! The code looks good, but look at my comment)

PR #1487, opened 2025-03-11
solution

DHlomozda:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (NotVaccinatedError,
```
Sezaro104, line 2: It's easier to import the entire file from the app.

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
```
Sezaro104, line 3: It's easier to import the entire file from the app.

app/main.py
```diff
@@ -1 +1,20 @@
...
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    number_mask = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
```
Sezaro104, line 12: You need to use the parent class VaccineError instead of listing the other two separated by commas.

Sezaro104 (approved):

PR #1518, opened 2025-03-30
Solution

olek007:


olek007:
@mateacademy-ai-mentor 
fixed type adnotation in is_outdated function

PR #1579, opened 2025-05-14
Solution

Nib20:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("Vaccine is outdated")
+        if not visitor.get("wearing_a_mask", False):
```
MNDonut, line 18: just a tip that you don't need `False` here as by default `get` returns `None` unless key is present. And `not None` will return `True`

MNDonut (approved):

PR #1638, opened 2025-07-04
create lockdown cafe

SerhiiBasok:


Hetmarum (approved):
круто прада фраза "Visitor must be mask to enter the cafe" дивна, "Visitor must **wear** mask to enter the cafe" було б ліпше. (in my humble opinion)

PR #1639, opened 2025-07-04
Solution

VadymKyrylko:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
+import datetime
```
yagginov: the import order is incorrect, you need
1.Standard library imports.
2.Related third party imports.
3.Local application/library specific imports.

It is also better not to write immediately after the open bracket, but to go to the next line:
```python
from app.main import (
    KnightWithArmour,
    KnightWithoutArmour,
    SuperHeroKnightWithGun
)
```

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def __str__(self) -> str:
+        return self.name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError
+        expiration_date = visitor["vaccine"].get("expiration_date")
+        if expiration_date is None:
+            raise NotVaccinatedError
```
yagginov: It seems that this second check is unnecessary here, plus you have code duplication (the same `raise` twice). If you want to check this, it's better to do it like this
```suggestion
        if (
            visitor.get("vaccine") is None 
            or visitor["vaccine"].get("expiration_date") is None
        ):
            raise NotVaccinatedError
```

app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+        if datetime.date.today() > expiration_date:
+            raise OutdatedVaccineError
+        if visitor["wearing_a_mask"] is False:
+            raise NotWearingMaskError(0)
+        else:
+            return f"Welcome to {self.name}"
```
yagginov: else is not needed here

yagginov (changes requested):
Overall good, just fix these small misstakes

yagginov (approved):
well

PR #1644, opened 2025-07-06
Implemented solution.

Otst0y:


app/errors.py
```diff
@@ -0,0 +1,14 @@
...
+class NotVaccinatedError(VaccineError):
+    pass
+
+
+class OutdatedVaccineError(VaccineError):
+    pass
+
+
+class NotWearingMaskError(Exception):
+    pass
```
Orixara, line 14: Maybe it will be great if you use docstring instead pass

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+    for visitor in friends:
+        try:
+            cafe.visit_cafe(visitor)
+        except VaccineError:
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+
+    if masks_to_buy != 0:
```
Orixara: And there in checklist we have good example with this, it will be great to change 
image

Orixara (changes requested):
All your code is great, but you have several small things to fix

Orixara (approved):
Great!

PR #1651, opened 2025-07-07
Solution

Vskesha:


app/cafe.py
```diff
@@ -0,0 +1,26 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor is not vaccinated")
+
+        expiration_date = visitor["vaccine"].get("expiration_date")
+
+        if not expiration_date or expiration_date < date.today():
+            raise OutdatedVaccineError("Visitor's vaccine is outdated")
```
Orixara, line 21: Here ai buddy marked me as a bug that I don't check that expiration_date is str and don't bring it to the date object, I realize that it's not required by the assignment but I think he will cuss

Orixara (approved):
Everything is clear I like your code style, I write about the same way, I marked you where most likely ai buddy will swear, have a nice day!

PR #1652, opened 2025-07-07
Develop

Artcrafterrra:


app/cafe.py
```diff
@@ -0,0 +1,24 @@
+import datetime
+from app.errors import (
```
reinvvay, line 2: You should add a blank line between imports of built-in libraries and imports of your project’s modules.

reinvvay (approved):

app/cafe.py
Artcrafterrra: True TNX

PR #1653, opened 2025-07-07
Solution

reinvvay:


app/main.py
```diff
@@ -1 +1,22 @@
...
+from app.cafe import Cafe
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    mask, vaccine = 0, 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            vaccine += 1
```
Artcrafterrra: It is not necessary to count how many vaccines are missing, since the absence of even one is enough to raise an error.

Artcrafterrra (approved):
LGTM

PR #1654, opened 2025-07-07
solution

Sashoxxx:


app/errors.py
```diff
@@ -0,0 +1,16 @@
+class VaccineError(Exception):
+    def __str__(self) -> str:
+        return "All friends should be vaccinated"
```
dimashevchukk: this msg is more about function visit_cafe rather than class VacineError
if you want to give a deffault msg to class, do it in init

app/errors.py
```diff
@@ -0,0 +1,16 @@
...
+    pass
+
+
+class OutdatedVaccineError(VaccineError):
+    pass
+
+
+class NotWearingMaskError(Exception):
+    def __str__(self) -> str:
+        return "You should wear a mask"
```
dimashevchukk: same here

app/main.py
```diff
@@ -1 +1,20 @@
...
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError as error:
+            return str(error)
+        except NotWearingMaskError:
+            count_masks += 1
+
+    if count_masks > 0:
+        return f"Friends should buy {count_masks} masks"
+    else:
```
dimashevchukk: not needed else

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from typing import List, Union
+
+from app.cafe import Cafe
+from app.errors import VaccineError, NotWearingMaskError
+
+
+def go_to_cafe(friends: Union[List[dict], dict], cafe: Cafe) -> str:
```
dimashevchukk: method works with list friends, so using Union is wrong

dimashevchukk (changes requested):

app/errors.py
```diff
@@ -0,0 +1,16 @@
+class VaccineError(Exception):
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
```
dimashevchukk, line 3: not necessary

app/errors.py
```diff
@@ -0,0 +1,16 @@
...
+    pass
+
+
+class OutdatedVaccineError(VaccineError):
+    pass
+
+
+class NotWearingMaskError(Exception):
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
```
dimashevchukk, line 16: not necessary

app/cafe.py
```diff
@@ -0,0 +1,29 @@
...
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("All friends should be vaccinated")
+        if not self.is_valid_vaccine_date(
+                visitor["vaccine"]["expiration_date"]
+        ):
+            raise OutdatedVaccineError("All friends should be vaccinated")
+        if visitor["wearing_a_mask"] is False:
```
dimashevchukk, line 22: get would be safer

dimashevchukk (approved):

PR #1657, opened 2025-07-07
solution

MorsImmortalis05:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("Visitor not vaccinated")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            raise OutdatedVaccineError("Vaccine outdated")
```
Plovuezz, line 17: Sorry, i have to reject 👉👈

You must fix this (Create a `VaccineError` parent class for `NotVaccinatedError` and `OutdatedVaccineError` errors.
Use it to catch both types of errors in the same `except` clause.)

Plovuezz (changes requested):

app/cafe.py
MorsImmortalis05: sorry Bohdan said you're wrong(
I have VaccineError in main.py and you looked in the cafe.py, there shouldn't be VaccineError

Plovuezz (approved):
💀💀💀

PR #1661, opened 2025-07-07
Sloution

MakovM:


app/cafe.py
```diff
@@ -0,0 +1,29 @@
...
+        name = visitor["name"]
+
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError(f"{name} is not vaccinated")
+
+        vaccine = visitor["vaccine"]
+        expiration_date = vaccine["expiration_date"]
+
+        if expiration_date < date.today():
+            raise OutdatedVaccineError(f"{name}'s vaccine is expired")
```
yagginov: if the expiration_date key is not present, the program will terminate its execution (Aishka is complaining about this in this task, so it will still need to be fixed)
```suggestion
        expiration_date = vaccine.get("expiration_date", date(2000, 1, 1)) # if there is no date, then we take the default as knowingly overdue

        if expiration_date < date.today():
            raise OutdatedVaccineError(f"{name}'s vaccine is expired")
```

app/cafe.py
```diff
@@ -0,0 +1,29 @@
...
+
+        if not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError(f"{name} is not wearing a mask")
```
yagginov: it is safer to retrieve such keys via the get method

app/main.py
```diff
@@ -1 +1,20 @@
...
+)
+
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            return "All friends should be vaccinated"
```
yagginov:
```suggestion
        except VaccineError:
            return "All friends should be vaccinated"
```

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
```
yagginov:
```suggestion
from app.errors import VaccineError, NotWearingMaskError
```

app/main.py
```diff
@@ -1 +1,20 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except (NotVaccinatedError, OutdatedVaccineError):
+            return "All friends should be vaccinated"
+        except NotWearingMaskError:
+            masks_to_buy += 1
+    if masks_to_buy > 0:
+        return f"Friends should buy {masks_to_buy} masks"
+    return f"Friends can go to {cafe.name}"
```
yagginov:
```suggestion
    if masks_to_buy:
        return f"Friends should buy {masks_to_buy} masks"
    return f"Friends can go to {cafe.name}"
```

yagginov (changes requested):
only small mistakes

yagginov (approved):
good

PR #1667, opened 2025-07-08
Solution

DenisPrudius:


app/cafe.py
```diff
@@ -0,0 +1,21 @@
+import datetime
```
trellle, line 1: you can write:
`from datetime import date`

app/cafe.py
```diff
@@ -0,0 +1,21 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        self.visitor = visitor
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError()
+        expiration_date = visitor["vaccine"].get("expiration_date")
+        if expiration_date < datetime.date.today():
```
trellle, line 18: and use date.today() instead

trellle (approved):
👍

PR #1744, opened 2025-09-06
cafe-implementation: implemented cafe

kenu21:


kenu21:
@mateacademy-ai-mentor 

> Style: Ensure consistent use of double quotes across this file, and that the file ends with exactly one blank line (checklist style requirement).

If I follow your advice I'll get error with flake8!

> Lines A-B: Remove print statements that output cafe.visit_cafe(friend) results and exception messages. The checklist doesn’t allow printing here; only return the specified strings.

Then why I should provide error message if we don't use it? I don't think your advice is good

kenu21:
@mateacademy-ai-mentor 

Printing does not create side effects and does not cause test failures.
I return only what is required. Everything else is printed to the console. I print exceptions because the tests require passing the messages to them. If these messages are not printed, their existence is meaningless, and in the tests these messages are mandatory!

PR #1752, opened 2025-09-11
Solution

Alex-x000:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+            number += 1
+            raise NotWearingMaskError(f"Visitor {visitor["name"]} "
+                                      f"is not wearing a mask.")
+        if "vaccine" not in visitor.keys():
+            number += 1
+            raise NotVaccinatedError(f"Visitor {visitor["name"]} "
+                                     f"is not vaccinated.")
+        elif visitor["vaccine"]["expiration_date"] < datetime.date.today():
+            number += 1
+            raise OutdatedVaccineError(f"Visitor {visitor["name"]} "
```
Dimosphen1: In case you use quotes inside of other quotes, inner quotes should not contradict the outer ones. So you should use
```suggestion
            raise OutdatedVaccineError(f"Visitor {visitor['name']} "

```

Dimosphen1 (changes requested):
Several changes were requested besides the changes from AI Buddy to fix flake8

app/cafe.py
```diff
@@ -0,0 +1,22 @@
+import datetime
+
+from app.errors import (NotWearingMaskError,
+                        NotVaccinatedError, OutdatedVaccineError)
```
Dimosphen1, line 4: Place imports on a new line and in an alphabetical order in the future, as follows
```suggestion
from app.errors import (
    OutdatedVaccineError,
    NotVaccinatedError,
    NotWearingMaskError,
)

```

Dimosphen1 (approved):
Looks good to me, well done! One comment is left for consideration

app/cafe.py
Alex-x000: thanks

PR #1785, opened 2025-10-06
Solution

henriquejhansen:


app/cafe.py
meruyme, line 1: Henrique, your code is correct, so I'll approve your PR! But if possible please implement the suggestions of our AI Mentor, it's a good practice in Python to end all files with a single blank line.

meruyme (approved):

PR #1823, opened 2025-10-29
'Solution'

AnastasiaCibart:


app/main.py
```diff
@@ -1 +1,18 @@
...
+            return "All friends should be vaccinated"
+        expiration = vaccine.get("expiration_date")
+        if datetime.date.today() > expiration:
+            return "All friends should be vaccinated"
+
+    masks_to_buy = sum(1 for f in friends if not f.get("wearing_a_mask"))
+    if masks_to_buy > 0:
+        return f"Friends should buy {masks_to_buy} masks"
+
+    return f"Friends can go to {cafe.name}"
```
vsmutok: This is not the best implementation, since you’ve already done it in `visit_cafe`.
Here, you just need to check whether the appropriate exception is raised or not.

app/main.py
```diff
@@ -1 +1,18 @@
...
+    for friend in friends:
+        vaccine = friend.get("vaccine")
+        if not vaccine:
+            return "All friends should be vaccinated"
+        expiration = vaccine.get("expiration_date")
+        if datetime.date.today() > expiration:
+            return "All friends should be vaccinated"
+
+    masks_to_buy = sum(1 for f in friends if not f.get("wearing_a_mask"))
+    if masks_to_buy > 0:
```
vsmutok: Don’t use the loop twice — do it only once.

vsmutok (changes requested):

DaTrEvTeR (approved):
Well done!

PR #1833, opened 2025-11-03
Solution

leemonch1k-space:


Vorongor (approved):
Your solution is correct and clean.
The code is clean, readable, and uses custom exceptions effectively to model real-world access rules. Responsibilities are well-split between validation inside the Cafe class and group-level handling in go_to_cafe. Error handling logic works but could be slightly improved by flattening conditional nesting and validating visitor input more defensively.  Overall, strong structure with minor refinements for robustness and clarity.

PR #1836, opened 2025-11-04
add solution

VadymVynokur:


app/cafe.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError(
+                "Visitor must have a valid vaccine.")
+
+        if visitor["vaccine"]["expiration_date"] < date.today():
```
stanislav-kishechko, line 20: vaccine can be put into a separate variable

stanislav-kishechko, line 20: It is better to get this parameter through the method get

app/cafe.py
```diff
@@ -0,0 +1,28 @@
...
+            raise OutdatedVaccineError(
+                "Visitor's vaccine has expired.")
+
+        if not visitor["wearing_a_mask"]:
```
stanislav-kishechko, line 24: It is better to get this parameter through the method get

app/main.py
```diff
@@ -1 +1,18 @@
...
+            cafe.visit_cafe(friend)
+
+        except VaccineError:
+            return "All friends should be vaccinated"
+
+        except NotWearingMaskError:
+            count += 1
+
+    return f"Friends should buy {count} masks" if count\
+        else f"Friends can go to {cafe.name}"
```
stanislav-kishechko, line 18: In this case, the ternary operator worsens readability.

stanislav-kishechko (approved):

PR #1840, opened 2025-11-05
Solution

vakt159:


app/cafe.py
```diff
@@ -0,0 +1,20 @@
...
+    OutdatedVaccineError,
+    NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine", False):
```
ostapshpeha: Why False? "not" is enough

app/main.py
```diff
@@ -1 +1,20 @@
...
+
+def go_to_cafe(friends: list, cafe: Cafe) -> str:
+    masks_to_buy = 0
+    correct_count = 0
+    for friend in friends:
+
+        try:
+            cafe.visit_cafe(friend)
+            correct_count += 1
+        except NotWearingMaskError:
```
ostapshpeha: You should catch VaccinatedError first due to checklist

ostapshpeha (changes requested):

ostapshpeha (approved):

PR #1843, opened 2025-11-06
py-lockdown-cafe-solution

oleksii-shcherbak-code:


app/errors.py
```diff
@@ -0,0 +1,17 @@
+class VaccineError(Exception):
+    pass
+
+
+class NotVaccinatedError(VaccineError):
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
```
vakt159: don't think it's necessary to create __init__ method

app/errors.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class OutdatedVaccineError(VaccineError):
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
```
vakt159: and here

app/errors.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class NotWearingMaskError(Exception):
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
```
vakt159: and here

vakt159 (changes requested):
remove useless __init__  methods  in exception classes and everything will be good

vakt159 (approved):

PR #1846, opened 2025-11-06
Solution

Ann-Riabykina:


app/cafe.py
```diff
@@ -0,0 +1,23 @@
+import datetime
+from typing import Dict
+from .errors import (NotVaccinatedError,
+                     OutdatedVaccineError, NotWearingMaskError)
```
AngryAndrii: if you have stated write each class on new line, please continue

app/cafe.py
```diff
@@ -0,0 +1,23 @@
+import datetime
+from typing import Dict
+from .errors import (NotVaccinatedError,
+                     OutdatedVaccineError, NotWearingMaskError)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name: str = name
```
AngryAndrii: Do we need to write type here, if we already have it on init arguments?

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from typing import List, Dict
+from app.cafe import Cafe
```
AngryAndrii, line 3: Import of standart libraries and local modules must be separated with one blank line. Please read checklist

app/main.py
```diff
@@ -1 +1,18 @@
...
+from app.errors import VaccineError, NotWearingMaskError
```
AngryAndrii: Import of several classes must be in scopes, for example:
`from app.main import (
    KnightWithArmour,
    KnightWithoutArmour,
    SuperHeroKnightWithGun
)`
please, read the checklist

AngryAndrii (changes requested):

AngryAndrii (approved):

PR #1856, opened 2025-11-07
Solution

Guillelmina8:


app/cafe.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        all_ok = True
+        if "vaccine" not in visitor:
+            all_ok = False
+            raise NotVaccinatedError(f"Visitor {visitor["name"]} "
```
tradewcs: Don't use nested double quotes. Change outer quotes to single quotes

tradewcs (changes requested):

app/cafe.py
Guillelmina8: I already removed all additional info such as visitor's name to make less problems with quotes.

PR #1886, opened 2025-12-03
Cafe class and corresponding errors for validation in functions are i…

ivanlavrinenko:
…mplemented

### DIFF
--- a/app/cafe.py
+++ b/app/cafe.py
@@ -1,8 +1,10 @@
 from datetime import date
 
-from app.errors import (NotVaccinatedError,
-                        OutdatedVaccineError,
-                        NotWearingMaskError)
+from app.errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
 
 
 class Cafe:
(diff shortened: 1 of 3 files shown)


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("Visitor is not vaccinated")
+        if visitor["vaccine"]["expiration_date"] < date.today():
+            raise OutdatedVaccineError("Visitor has outdated vaccine")
+        if visitor["wearing_a_mask"] is False:
+            raise NotWearingMaskError("Visitor nis not wearing mask")
```
vsmutok: There's a small typo in the error message. It should be "is not" instead of "nis not".

app/cafe.py
```diff
@@ -0,0 +1,19 @@
+from datetime import date
+
+from app.errors import (NotVaccinatedError,
+                        OutdatedVaccineError,
+                        NotWearingMaskError)
```
vsmutok:
```suggestion
from app.errors import (
    NotVaccinatedError,
    OutdatedVaccineError,
    NotWearingMaskError
)
```

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #2021, opened 2026-03-02
Implement Class Cafe. Add custom errros. Implement logic try exept

HalasOleh:


HalasOleh (commented):
123

app/cafe.py
```diff
@@ -0,0 +1,35 @@
+import datetime
+from .errors import (
+    NotVaccinatedError,
+    OutdatedVaccineError,
+    NotWearingMaskError,
+)
```
vsmutok: This violates checklist item [#8](https://github.com/mate-academy/py-lockdown-cafe/pull/8): 'Use absolute imports instead of relative imports'. You should use an absolute import like `from app.errors import ...`.

vsmutok (changes requested):

Boyok1337 (approved):
LGTM!

PR #2082, opened 2026-03-25
Develop

PavloDanylkiv:


app/main.py
```diff
@@ -1 +1,20 @@
...
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+            counter += 1
+        except VaccineError as e:
+            return f"{e}"
+        except NotWearingMaskError:
+            wear_mask += 1
+    if len(friends) == counter:
+        return f"Friends can go to {cafe.name}"
```
vsmutok, line 17: This check is unnecessary here.

app/cafe.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def visit_cafe(self, visitor: dict) -> None | str:
+        today = date.today()
+        if "vaccine" not in visitor:
+            raise NotVaccinatedError("All friends should be vaccinated")
+        elif today > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("All friends should be vaccinated")
+        elif not visitor["wearing_a_mask"]:
+            raise NotWearingMaskError("Error")
+        else:
```
vsmutok, line 21: no need to use `else` here.

vsmutok (changes requested):

app/main.py
PavloDanylkiv: @vsmutok I don't really understand how to do without this check?

PR #2109, opened 2026-04-04
'Solution'

artradkevych:


app/cafe.py
```diff
@@ -0,0 +1,19 @@
...
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        if not visitor.get("vaccine"):
+            raise NotVaccinatedError("You are not vaccinated")
+        if datetime.date.today() > visitor["vaccine"]["expiration_date"]:
+            raise OutdatedVaccineError("Your vaccine is outdated")
+        if not visitor.get("wearing_a_mask", False):
```
petrykivd, line 17: Try to put the error messages inside the exception classes, not in the raise line.
Two options:

Override 'init' with a default arg: 'def init(self, message: str = "You are not vaccinated") -> None' — flexible, still allows custom messages.
Override 'str': 'def str(self) -> str: return "You are not vaccinated"'— simpler, but ignores any message passed to the constructor.
Option 1 is generally preferred 😉

petrykivd (approved):
Good job! 🚀
Keep these comment in mind for future tasks.

PR #2127, opened 2026-04-10
all steps of develop

andriihrechko:


alisahoys (approved):
Clear exception hierarchy and separation of concerns

PR #2131, opened 2026-04-12
Implemented logic in `main.py`, `errors.py`, `cafe.py` modules

zwirok-git:


dimaaa-cmd (approved):

app/cafe.py
```diff
@@ -0,0 +1,35 @@
...
+    OutdatedVaccineError,
+    NotWearingMaskError
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.__name = name
+
+    @property
```
PavelPereverzev1: I don't see any need to use @property here, since the attribute value isn't modified from the outside ( setter not implemented).

app/cafe.py
```diff
@@ -0,0 +1,35 @@
...
+    def success_message(self) -> str:
+        return f"Welcome to {self.__name}"
+
+    @property
```
PavelPereverzev1, line 14: Since this value is initialized once and doesn't change from the outside, we don't really need @property

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+
+def go_to_cafe(friends: list[dict], cafe: Cafe) -> str:
+    masks_to_buy = 0
+
+    for friend in friends:
+        try:
+            cafe.visit_cafe(friend)
+        except VaccineError:
+            return "All friends should be vaccinated"
```
PavelPereverzev1, line 15: The current logic breaks the function as soon as it hits an unvaccinated friend. This means the total mask count is lost, which contradicts the requirement to calculate all masks for the group.

PavelPereverzev1 (changes requested):

app/main.py
zwirok-git: Your claim is irrelevant. Read more carefully README.md

app/cafe.py
zwirok-git: it's read-only interface

PR #2136, opened 2026-04-13
Implementation of the Сafe class

amesemyta1:


app/main.py
ratatuiii, line 1: OutdatedVaccineError is never mentioned in the File

ratatuiii (approved):
Good to go

PR #2138, opened 2026-04-13
Implement Cafe class with vaccine and mask validation, add go_to_cafe…

Prus-Oleksandr:
… function with error handling


vova-dev-python (approved):
Great implementation! The logic in main.py is very robust, especially how you handle the priority of vaccine checks over mask counts. Here are a few points: Remember Checklist Item #6. Instead of list and dict, use more specific types like list[dict]. In cafe.py, accessing visitor["wearing_a_mask"] directly might cause a KeyError if the key is missing.

PR #2257, opened 2026-07-04
Solution

Teashrock:


Teashrock:
@mateacademy-ai-mentor 
As for checklist item [#3](https://github.com/mate-academy/py-lockdown-cafe/pull/3): I've fixed it for `NotWearingMaskError`, as it also allows to pass visitor's name easily. However, regarding `VaccineError`-derived classes, I decided to implement their detailed messages with their own `__str__` methods, utilising `visitor_name` variable passed to `VaccineError.__init__`. So now, both `NotVaccinatedError` and `OutdatedVaccineError` return detailed messages as intended, just through their own `__str__`s rather than direct input upon `raise`. Is this really not a correct approach? Should I remove `__str__` methods completely, and implement their messages upon `raise`?

Teashrock:
@mateacademy-ai-mentor 
Done. As you recommended, I've reimplemented `NotWearingMaskError` the same way as both `VaccineErrors`. Now it's called the same exact way as them. I've also deleted the redundant `print(err)` message, as well as the unused variable `err`.

vsmutok (approved):
Great!

PR #2259, opened 2026-07-07
lockdown-cafe 0.1

EldarRasulov:


hladiivolodymyr (changes requested):
Use absolute imports instead of relative imports.

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+from .cafe import Cafe
+from .errors import VaccineError, NotWearingMaskError
```
vsmutok: Use absolute imports instead of relative imports

app/cafe.py
```diff
@@ -0,0 +1,24 @@
+from datetime import date
+
+from .errors import (
```
vsmutok: Use absolute imports instead of relative imports

vsmutok (changes requested):

app/cafe.py
```diff
@@ -0,0 +1,24 @@
...
+    OutdatedVaccineError,
+    NotWearingMaskError,
+)
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
```
Sn1kls, line 14: optional, for the future: you can type `visitor` more precisely as `dict[str, Any]` (needs import `Any`)

Sn1kls (approved):
Approved. Left one small optional note inline for the future

PR #2305, opened 2026-08-04
Solution

gospodenkolubomir-oss:


app/cafe.py
```diff
@@ -0,0 +1,18 @@
+from app import errors
+import datetime
+
+
+class Cafe:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def visit_cafe(self, visitor: dict) -> str:
+        # 1. Перевірка на вакцину
```
Sn1kls, line 10: It can be removed or translated to English

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from app.cafe import Cafe
+from app.errors import NotWearingMaskError, VaccineError
+
+
+def go_to_cafe(friends: str, cafe: Cafe) -> str:
```
Sn1kls, line 5: `friends` is a list of visitor dicts, so it should be friends: `list[dict]`, not `str`

app/cafe.py
```diff
@@ -0,0 +1,18 @@
+from app import errors
+import datetime
```
Sn1kls, line 2: standard-library imports come first, so put i`mport datetime` above `from app import errors` (ideally with a blank line between the two groups)

Sn1kls (changes requested):

